PR #15, opened 2012-09-07
don't validate the issue id if no jira password is available

domruf:
this is a quick fix for JENKINS-9549/JENKINS-1904


warden:
This was fixed in #55 . Thanks!

PR #19, opened 2013-03-05
Updating individual jira issues should not fail the build but rather log a warning message and continue

igorekbsu:
We are using jira-plugin as a post build step to mark all the relevant issues with the release version. However, if some of the issues had been marked as closed, the pluging logs a nasty message saying that the issue can not be updated (because it is closed) and fails the build. 
The proposed changes deal with this problem by logging a warning message if a particular issue can not be updated and continue marking other issues.


warden:
I actually agree with both usecases.
The main role of the plugin is to update the issues but this is not a build plugin, but a build-reference plugin. So as far as my understanding of workflow is concerned, this should not fail builds by default. 

That being said, because it already does... I think we could introduce a switch "fail build when updating JIRA fails", that is on by default.

[JENKINS-28879] (https://issues.jenkins-ci.org/browse/JENKINS-28879)

PR #20, opened 2013-03-28
Adds jira-url replacement parameter

mavlyutov:
Hi!
In my company robots(and also jenkins ci) has access only to read-only jira-instances.
So in changes frame jenkins shows me wrong links.
I added a parameter in main config to be able to replace real jira url to fake and vise versa.
Check it please


warden:
Functionality added in #21 .

PR #27, opened 2013-06-12
[FIXED JENKINS-3709, JENKINS-18316] Protect JiraSite password using hudson.util.Secret and validate Jira site Link URL

tyrius02:
JENKINS-3709: Passwords used for the JiraSite were stored in the clear and also sent over the wire in the clear.  By wrapping the password with hudson.util.Secret the password is obscured both when persisted and sent over the wire.

JENKINS-18316: The Jira site Link URL is not being validated even though it is mandatory.  If the field is left blank when the Validate button is clicked a MalformedURLException will be thrown and the user will not be informed why validation failed.

### DIFF
--- a/src/main/java/hudson/plugins/jira/JiraSite.java
+++ b/src/main/java/hudson/plugins/jira/JiraSite.java
@@ -650,6 +650,9 @@ public class JiraSite extends AbstractDescribableImpl<JiraSite> {
             if (url == null) {// URL not entered yet
                 return FormValidation.error("No URL given");
             }
+            if (alternativeUrl == null) { // Link URL not entered yet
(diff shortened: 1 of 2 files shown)


AndreaGiardini:
This security fix is unmerged since 2013?

warden:
For me this worked as well, so part for JENKINS-3709 was merged. I left out the alternativeUrl part as this seems an optional field.

AndreaGiardini:
What does it mean? all the password are still stored in cleartext

warden:
The password will be stored in encrypted format when JIRA configuration is updated. Currently there is no logic added to migrate the old passwords automatically.

PR #31, opened 2013-08-13
Use the first release version as default value.

chrislovsund:
[JENKINS-19178] - Use default Release Version Parameter when parameterized build is triggered
Using the first in the list of release versions received from JIRA as default value.


warden:
The code for sorting JIRA versions is now in master, which makes this PR obsolete because:
* the order of versions might have changed
* this can be resolved via parameter passing jiraRelease/jiraProjectKey variables or by version creation as @artkoshelev suggested. Discussion is welcome though on how to generalize this in the UI worklow.
One of the options would be to add a "Use newest version" option to JiraReleaseVersionUpdaterBuilder and getNewestVersion to JiraSite.

PR #50, opened 2015-03-27
JENKINS-17156: remove issue from carry over list after submitting the co...

krulls:
Issues [JENKINS-17156] (https://issues.jenkins-ci.org/browse/JENKINS-17156)  and [JENKINS-13500] (https://issues.jenkins-ci.org/browse/JENKINS-13500) show that when submitting comments to JIRA issues the successfully commented issues won't be removed from the carry over list when only one RemoteException happens. 

This request separates comment adding for each issue and keeps only relevant issues in the carry over list.


src/main/java/hudson/plugins/jira/Updater.java
```diff
@@ -127,22 +127,38 @@ static void submitComments(
...
                         createComment(build, useWikiStyleComments, jenkinsRootUrl, recordScmChanges, issue),
                         groupVisibility, roleVisibility);
-            } catch (RemotePermissionException e) {
-                // Seems like RemotePermissionException can mean 'no permission' as well as
-                // 'issue doesn't exist'.
-                // To prevent carrying forward invalid issues forever, we have to drop them
-                // even if the cause of the exception was different.
-                logger.println("Looks like " + issue.id + " is no valid JIRA issue or you don't have permission to update the issue.\n" +
-                        "Issue will not be updated.\n" + e);
+                } catch (RemotePermissionException e) {
```
artkoshelev: shouldn't we keep such issues as well?

src/main/java/hudson/plugins/jira/Updater.java
```diff
@@ -127,22 +127,38 @@ static void submitComments(
...
+                    // Seems like RemotePermissionException can mean 'no permission' as well as
+                    // 'issue doesn't exist'.
+                    // To prevent carrying forward invalid issues forever, we have to drop them
+                    // even if the cause of the exception was different.
+                    logger.println("Looks like " + issue.id + " is no valid JIRA issue or you don't have permission to update the issue.\n" +
+                            "Issue will not be updated.\n" + e);
+                } catch (RemoteException e) {
+                	if( 0 < issuesWithExceptions.length()) {
```
artkoshelev: please switch this comparison e.g. x > 0

src/main/java/hudson/plugins/jira/Updater.java
```diff
@@ -127,22 +127,38 @@ static void submitComments(
...
+                	logger.println(Messages.Updater_FailedToCommentOnIssue(issue.id));
+                	continue;
+                }
+                
                 issues.remove(issue);
-            }
+        }
+        
+        //	throw RemoteException if problems occured
+        if( 0 < issuesWithExceptions.length() ) {
```
artkoshelev: and this one

artkoshelev:
ILGTM overall, just a few bits need to be fixed. Please check the code formatting, it seems sometimes you use spaces and sometimes tabs, we use only spaces.

src/main/java/hudson/plugins/jira/Updater.java
artkoshelev: and according to codestyle it should look like **if (x > 0)**, please fix it in other places as well

src/test/java/hudson/plugins/jira/UpdaterTest.java
```diff
@@ -352,5 +355,75 @@ public void testUserPatternMatchTwoIssuesInOneComment() {
         Assert.assertTrue(ids.contains("FOOBAR-4711"));
         Assert.assertTrue(ids.contains("FOOBAR-21"));
     }
+    
+    @Test
+    @Bug(17156)
+    public void testIssueIsRemovedFromCarryOverListAfterSubmission() throws RemoteException {
+    	//	create CarryOverList
```
artkoshelev: use multiline comment style /** */

src/test/java/hudson/plugins/jira/UpdaterTest.java
```diff
@@ -352,5 +355,75 @@ public void testUserPatternMatchTwoIssuesInOneComment() {
...
+        ChangeLogSet changeLogSet = ChangeLogSet.createEmpty(build);
+        when(build.getChangeSet()).thenReturn(changeLogSet);
+        when(build.getResult()).thenReturn(Result.SUCCESS);
+        
+        final String groupVisibility = "";
+        final String roleVisibility = "";
+        
+        try {
+        	Updater.submitComments(build,
+                System.out, "http://jenkins", issues, session, false, false, groupVisibility, roleVisibility);
```
artkoshelev: you should add fail() after this line, because you expect to be in a catch block already

src/test/java/hudson/plugins/jira/UpdaterTest.java
```diff
@@ -352,5 +355,75 @@ public void testUserPatternMatchTwoIssuesInOneComment() {
...
+        } catch (RemoteException e)
+        {
+        	// expected
+        }
+        
+        // expected issue list
+        final List<JiraIssue> expectedIssuesToCarryOver = Lists.newArrayList(brokenIssue);
+
+        Assert.assertThat(issues, CoreMatchers.is(expectedIssuesToCarryOver));
```
artkoshelev: let's move assertion in a catch block, because it's expected flow

artkoshelev: and made please a static import for **is** matcher

src/main/java/hudson/plugins/jira/Updater.java
krulls: It depends. In my opinion a configuration switch is needed for the decision. The current implementation comes from contributor 'kutzi', a commit with hash '2c146187a1cdf...'. I just took over.

krulls:
fixed the review issues. for the record: could you please dissolve 'ILGTM'? thanks ;)

PR #51, opened 2015-04-03
[JENKINS-23292] Add system property to disable jira mail address resolving

christ66:
Setting the system property `-Dhudson.plugins.jira.JiraMailAddressResolver.DISABLE=true` will disable resolving email addresses with Jira.

The reason for disabling the JiraMailAddressResolver is because it will access Jira to obtain the email address which could potentially return an invalid email (as in `JENKINS-23292`). Disabling the email resolver for Jira will allow the email addresses to be resolved through another method (possibly with `LDAP`).

### DIFF
--- a/src/main/java/hudson/plugins/jira/JiraMailAddressResolver.java
+++ b/src/main/java/hudson/plugins/jira/JiraMailAddressResolver.java
@@ -25,7 +25,7 @@ public class JiraMailAddressResolver extends MailAddressResolver {
      *
      * To disable set the System property "-Dhudson.plugins.jira.JiraMailAddressResolver.DISABLE=true"
      */
-    public static final boolean DISABLE = Boolean.getBoolean(JiraMailAddressResolver.class.getName() + ".DISABLE");
+    public static boolean DISABLE = Boolean.getBoolean(JiraMailAddressResolver.class.getName() + ".DISABLE");
 
     @Override
     public String findMailAddressFor(User u) {
--- a/src/main/java/hudson/plugins/jira/JiraMailAddressResolver.java
+++ b/src/main/java/hudson/plugins/jira/JiraMailAddressResolver.java
@@ -20,8 +20,18 @@ import java.util.regex.Pattern;
 public class JiraMailAddressResolver extends MailAddressResolver {
     private static final Logger LOGGER = Logger.getLogger(JiraMailAddressResolver.class.getName());
 
+    /**
+     * Boolean to disable the Jira mail address resolver.
+     *
+     * To disable set the System property "-Dhudson.plugins.jira.JiraMailAddressResolver.DISABLE=true"
+     */
+    public static final boolean DISABLE = Boolean.getBoolean(JiraMailAddressResolver.class.getName() + ".DISABLE");
+
     @Override
     public String findMailAddressFor(User u) {
+        if (DISABLE)
+            return null;
+
         String username = u.getId();
 
         for (JiraSite site : JiraProjectProperty.DESCRIPTOR.getSites()) {
--- a/src/main/java/hudson/plugins/jira/JiraMailAddressResolver.java
+++ b/src/main/java/hudson/plugins/jira/JiraMailAddressResolver.java
@@ -23,13 +23,13 @@ public class JiraMailAddressResolver extends MailAddressResolver {
     /**
      * Boolean to disable the Jira mail address resolver.
      *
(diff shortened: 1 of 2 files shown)


olamy:
Sounds good.
But what about a configuration via the ui?

daniel-beck:
This is probably an unusual enough use case to not warrant a UI option.

src/main/java/hudson/plugins/jira/JiraMailAddressResolver.java
```diff
@@ -20,8 +20,18 @@
 public class JiraMailAddressResolver extends MailAddressResolver {
     private static final Logger LOGGER = Logger.getLogger(JiraMailAddressResolver.class.getName());
 
+    /**
+     * Boolean to disable the Jira mail address resolver.
+     *
+     * To disable set the System property "-Dhudson.plugins.jira.JiraMailAddressResolver.DISABLE=true"
+     */
+    public static final boolean DISABLE = Boolean.getBoolean(JiraMailAddressResolver.class.getName() + ".DISABLE");
```
daniel-beck: Why is this `final`?

walterk82:
What's the stopper on this pull request?

artkoshelev:
I don't think using class name in system property is a good idea. In case you rename this class, it will automatically change property name you read, so it will break functionality.

And please provide tests for your changes.

src/main/java/hudson/plugins/jira/JiraMailAddressResolver.java
christ66: Since we are setting this argument through command line it shouldn't be modifiable. If I remove the `final`, then I can use a groovy script to set the value, and makes the argument useless.

daniel-beck: > then I can use a groovy script to set the value

Exactly. You'll be able to test whether this has the desired effect while troubleshooting without having to reboot the production instance.

daniel-beck:
> I don't think using class name in system property is a good idea.

This is an often-used pattern in Jenkins, see https://wiki.jenkins-ci.org/display/JENKINS/Features+controlled+by+system+properties

Backwards compatibility can easily be achieved by moving to a fixed string when renaming.

Note that plugins that depend on Jira Plugin today may also break in case you rename the class, as it is a public class.
https://wiki.jenkins-ci.org/display/JENKINS/Governance+Document#GovernanceDocument-Compatibilitymatters

src/main/java/hudson/plugins/jira/JiraMailAddressResolver.java
christ66: Fair enough. I'll remove `final`.

christ66:
Test has been added.

src/test/java/hudson/plugins/jira/MailResolverTest.java
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * @author schristou88
+ */
+public class MailResolverTest extends JenkinsRule {
+    @Rule
+    public JenkinsRule r = new JenkinsRule();
+
+    @BeforeClass
+    public static void setUp() throws Exception {
+        System.setProperty(JiraMailAddressResolver.class.getName() + ".DISABLE", "true");
```
daniel-beck: If this used the full class name in a fixed string, it'd protect against class rename/move.

src/main/java/hudson/plugins/jira/JiraMailAddressResolver.java
```diff
@@ -20,8 +20,18 @@
 public class JiraMailAddressResolver extends MailAddressResolver {
     private static final Logger LOGGER = Logger.getLogger(JiraMailAddressResolver.class.getName());
 
+    /**
+     * Boolean to disable the Jira mail address resolver.
+     *
+     * To disable set the System property "-Dhudson.plugins.jira.JiraMailAddressResolver.DISABLE=true"
+     */
+    public static boolean DISABLE = Boolean.getBoolean(JiraMailAddressResolver.class.getName() + ".DISABLE");
```
daniel-beck: Naming in all caps inconsistent with basically all system properties used by core.

walterk82:
Can we keep the progress moving on this?   Should this be named `JiraMailAddressResolver.class.getName() + ".disable"` or `JiraMailAddressResolver.class.getName() + ".disabled" `

Thanks.

artkoshelev:
".disabled" sounds good to me

walterk82:
I installed the snapshot and have confirmed that email resolving no longer occurs.  I think it would be safe to merge this change.

PR #52, opened 2015-04-08
Pluggable strategy of selecting which issues are relevant to update

FrantaM:
This sort of fixes [JENKINS-1691](https://issues.jenkins-ci.org/browse/JENKINS-1691) and [JENKINS-25014](https://issues.jenkins-ci.org/browse/JENKINS-25014) - it will require other plugins (probably custom ones) to provide issue selectors to get the demanded behavior.

### DIFF
--- a/src/main/java/hudson/plugins/jira/JiraIssueUpdater.java
+++ b/src/main/java/hudson/plugins/jira/JiraIssueUpdater.java
@@ -13,9 +13,11 @@ import hudson.tasks.BuildStepDescriptor;
 import hudson.tasks.BuildStepMonitor;
 import hudson.tasks.Publisher;
 import hudson.tasks.Recorder;
+import jenkins.model.Jenkins;
 import net.sf.json.JSONObject;
 import org.kohsuke.stapler.StaplerRequest;
 
+import org.kohsuke.stapler.DataBoundConstructor;
 import java.io.IOException;
 import java.io.PrintStream;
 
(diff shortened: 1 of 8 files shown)


olamy:
unit tests?

FrantaM:
I can add some for sure but what do you want me to test?

src/main/java/hudson/plugins/jira/Updater.java
```diff
@@ -341,4 +346,24 @@ static void findIssues(AbstractBuild<?, ?> build, Set<String> ids, Pattern patte
      * Debug flag.
      */
     public static boolean debug = false;
+
+    public static final class DefaultUpdaterIssuesSelector extends UpdaterIssuesSelector {
```
warden: Please move to a separate class file and check if that is covered by the existing unit tests. 
I guess it will be, but if not, please create a new one.

src/main/java/hudson/plugins/jira/JiraIssueUpdater.java
```diff
@@ -48,6 +55,12 @@ public DescriptorImpl getDescriptor() {
         return DESCRIPTOR;
     }
 
+    public UpdaterIssuesSelector getIssueSelector() {
```
warden: Please conform to one naming standard i.e. using singular or plurar form for classes and methods.

warden:
As you pointed out, this only makes it possible to fix the issues you mentioned. Do you have other implementations of the strategy that solve any of those two issues?

warden:
@FrantaM any chance you could resolve conflicts and fix the naming issues I mentioned?
For the simplest unit test I would propose DefaultUpdaterIssuesSelector logic.
Also, I would add some help message on the jelly script to say what does "Default selector" actually means.

PR #59, opened 2015-08-21
Progress workflow by commit logs

plateaukao:
Hello, 

Current jira-plugin only can update jira workflow status by assigning JQL queries. However, most of time, we would like to update jia issues that are mentioned in commit change logs. So, I modified a bit the code to allow this option.

Please let me know if it's okay to merge this change into next release. Thanks.

### DIFF
--- a/src/main/java/hudson/plugins/jira/JiraIssueUpdateBuilder.java
+++ b/src/main/java/hudson/plugins/jira/JiraIssueUpdateBuilder.java
@@ -31,6 +31,9 @@ import org.kohsuke.stapler.QueryParameter;
 
 import javax.servlet.ServletException;
 import java.io.IOException;
+import com.atlassian.jira.rest.client.api.domain.Issue;
+
+import java.util.List;
 
 /**
  * Build step that will mass-update all issues matching a JQL query, using the specified workflow
@@ -94,8 +97,22 @@ public class JiraIssueUpdateBuilder extends Builder {
 
         listener.getLogger().println("[JIRA] JQL: " + realJql);
 
+            JiraSession session = null;
+            try {
+                session = site.createSession();
+            } catch (IOException e) {
+                listener.getLogger().println(Messages.Updater_FailedToConnect());
+                e.printStackTrace(listener.getLogger());
+            }
+            if (session == null) {
+                build.setResult(Result.FAILURE);
+                return true;
+            }
+
+	List<JiraIssue> issues = Updater.getJiraIssues(build, site, session, listener);
+
         try {
-            if (!site.progressMatchingIssues(realJql, realWorkflowActionName, realComment, listener.getLogger())) {
+            if (!site.progressIssues(issues, realWorkflowActionName, realComment, listener.getLogger())) {
                 listener.getLogger().println(Messages.JiraIssueUpdateBuilder_SomeIssuesFailed());
                 build.setResult(Result.UNSTABLE);
             }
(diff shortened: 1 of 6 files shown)


plateaukao:
@joelfire  you can check JiraIssueUpdateBuilder.java line 121 and 126; if the checkbox (value: updateRelatedIssues) is not checked, it will go the original JQL query way of updating status.

joelfire:
Ah, I see now in the later commit. I still don’t quite understand the use case though.

plateaukao:
While fixing issues in each commits, we write down jira issue number in commit logs. So, once the code is merged to the develop branch, and develop branch builds successfully, we would like to change these related jira issues' status to "Resolve" automatically, so that QA know it's time to verify these issues.

Is this description more clear for you?

src/main/java/hudson/plugins/jira/JiraIssueUpdateBuilder.java
```diff
@@ -94,11 +103,32 @@ public boolean perform(AbstractBuild<?, ?> build, Launcher launcher, BuildListen
 
         listener.getLogger().println("[JIRA] JQL: " + realJql);
 
+        JiraSession session = null;
         try {
-            if (!site.progressMatchingIssues(realJql, realWorkflowActionName, realComment, listener.getLogger())) {
-                listener.getLogger().println(Messages.JiraIssueUpdateBuilder_SomeIssuesFailed());
-                build.setResult(Result.UNSTABLE);
-            }
+            session = site.createSession();
```
warden: This method looks deprecated, did you use it in purpose?

src/main/resources/hudson/plugins/jira/JiraIssueUpdateBuilder/config.jelly
```diff
@@ -14,6 +14,9 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 <j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
+  <f:entry title="${%Update jira issues' status found in changelogs}" field="updateRelatedIssues">
+    <f:checkbox/>
```
warden: Looking at the options and on the logic, it is not clear when and when not jqlSearch is used. Apart from the fact that we should avoid nested if's, with this logic I think it would be better to use radio button instead of checkbox here as the jqlSearch will not be used in this case?

warden:
Daniel,
I've added some comments, I think this needs better description in the UI of how the logic behaves and/or usage of radio button instead of checkbox. Also, please squash the commits in the PR as they contain contradictory commits.

kralleman:
Just wanted to say that I have been looking for similar functionality in the original plugin. Hopefully you'll get the PR accepted.

warden:
I'll be happy to merge as long as the comments will get discussed/resolved :)

src/main/java/hudson/plugins/jira/JiraSite.java
```diff
@@ -594,6 +594,57 @@ public void migrateIssuesToFixVersion(String projectKey, String versionName, Str
         session.migrateIssuesToFixVersion(projectKey, versionName, query);
     }
 
+    public boolean progressIssues(List<JiraIssue> issueList, String workflowActionName, String comment, PrintStream console) throws IOException {
```
warden: I think it would be better to re-use/split the current progressMatchingIssues instead of almost duplicating it with another parameters.

warden:
I'll close this as #79 was merged and this should be now possible with JiraEnvironmentVariableBuilder.

PR #60, opened 2015-09-02
[FIXES JENKINS-30242] don't cast object in converter to string

lanwen:
https://issues.jenkins-ci.org/browse/JENKINS-30242

Instead use the toString method

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/jira-plugin/60)

### DIFF
--- a/src/main/java/hudson/plugins/jira/JiraProjectProperty.java
+++ b/src/main/java/hudson/plugins/jira/JiraProjectProperty.java
@@ -119,7 +119,7 @@ public class JiraProjectProperty extends JobProperty<AbstractProject<?, ?>> {
                         return null;
                     }
                     try {
-                        return new URL((String) o);
+                        return new URL(o.toString());
                     } catch (MalformedURLException e) {
                         LOGGER.warning(String.format("%s is not a valid URL.", o.toString()));
                         return null;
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/plugins/jira/JiraProjectProperty.java
```diff
@@ -113,25 +114,27 @@ public boolean configure(StaplerRequest req, JSONObject formData) {
...
-            }, java.net.URL.class);
+            Stapler.CONVERT_UTILS.register(new EmptyFriendlyURLConverter(), java.net.URL.class);
             //End hack
 
             sites.replaceBy(req.bindJSONToList(JiraSite.class, formData.get("sites")));
             save();
             return true;
         }
+
+        public static class EmptyFriendlyURLConverter implements Converter {
```
oleg-nenashev: NIT: Makes sense to make <code>@Restricted</code>. I doubt it should be exposed to the API, because it still smells like a hack

oleg-nenashev:
:+1: in general

src/main/java/hudson/plugins/jira/JiraProjectProperty.java
warden: Seems that @Restricted(DoNotUse) will not pass Jenkins build :(

lanwen: Replaced with `NoExternalUse`. Will see

lanwen:
thanks!

lanwen:
@warden hello, can you release this fix as a hotfix (as of many people wait for it)?
 Thanks!

PR #62, opened 2015-09-15
JENKINS-30333 use direct references instead of WeakReference, added c…

gerhard6:
…reateSession logging


src/main/java/hudson/plugins/jira/JiraSite.java
```diff
@@ -203,16 +201,11 @@ public String getName() {
      */
     @Nullable
```
artkoshelev: It seems it's not nullable any more?

gerhard6: It is still nullable, as the called createSession is also nullable. (on null user and password)

warden:
This was merged, but I think I lost the "Merged from PR #62 " commit message and that's why GH didn't catch this. I will close this for this reason now. Thanks for the fix!

PR #63, opened 2015-09-17
JiraCreateIssueNotifier tests and some refactoring

warden:


src/main/java/hudson/plugins/jira/JiraCreateIssueNotifier.java
```diff
@@ -96,7 +98,7 @@ public BuildStepMonitor getRequiredMonitorService() {
     @Override
     public boolean perform(AbstractBuild<?, ?> build, Launcher launcher, BuildListener listener) throws IOException, InterruptedException {
 
-        String jobDirPath = Jenkins.getInstance().getBuildDirFor(build.getProject()).getPath();
+        String jobDirPath = build.getProject().getBuildDir().getPath();
```
warden: @artkoshelev can you take a look at this? I started some refactoring but probably could use some help.
E.g. I'm not sure if this change I did is good, still trying to dig through all of this undocumented API... Any remarks about the changes or test optimizations are welcome.

src/main/java/hudson/plugins/jira/JiraCreateIssueNotifier.java
```diff
@@ -274,15 +282,10 @@ private JiraSession getJiraSession(AbstractBuild<?, ?> build) throws IOException
     /**
      * @param filename
      */
-    private void deleteFile(String filename, TaskListener listener) {
+    private void deleteFile(String filename) {
```
warden: Also I'm pretty sure that writing to the file is not the best way to preserve information between builds.. any hints this could be done in better way?

src/main/java/hudson/plugins/jira/JiraCreateIssueNotifier.java
```diff
@@ -134,25 +137,17 @@ public boolean perform(AbstractBuild<?, ?> build, Launcher launcher, BuildListen
...
-        EnvVars environmentVariable = build.getEnvironment(TaskListener.NULL);
-
-        String buildURL = environmentVariable.get("BUILD_URL");
-        String buildNumber = environmentVariable.get("BUILD_NUMBER");
-        String jobName = environmentVariable.get("JOB_NAME");
-        String jenkinsURL = Jenkins.getInstance().getRootUrl();
+        EnvVars vars = build.getEnvironment(TaskListener.NULL);
 
+        String summary = String.format("Build %s failed", getBuildName(vars));
         String checkDescription = (this.testDescription.equals("")) ? "No description is provided" : this.testDescription;
```
artkoshelev: we can probably get a NPE calling `testDescription.equals()` without checking if it's not null

artkoshelev: and what's the reason to call local variable checkDescription instead of testDescription? there is nothing but confuse

src/main/java/hudson/plugins/jira/JiraCreateIssueNotifier.java
artkoshelev: @oleg-nenashev can you help please with this?

src/main/java/hudson/plugins/jira/JiraCreateIssueNotifier.java
```diff
@@ -305,35 +300,30 @@ private void writeInFile(String Filename, Issue issue) throws FileNotFoundExcept
...
 
-                    //Status=1=Open OR Status=5=Resolved
-                    if (Status.equals("1") || Status.equals("5")) {
-                        listener.getLogger().println("The previous build also failed creating issue with issue ID " + issueId);
-                        addComment(build, issueId, comment);
-                    }
-
-                    if (Status.equals("6")) {
+                    // Issue Closed, need to open new one
+                    if (status.getName().equals("Closed") || status.getName().equals("Done")) {
```
artkoshelev: since statuses are pre-defined constants, it will be better to store them in Enum object

src/test/java/hudson/plugins/jira/JiraCreateIssueNotifierTest.java
```diff
@@ -0,0 +1,176 @@
+package hudson.plugins.jira;
+
+import com.atlassian.jira.rest.client.api.domain.Component;
+import com.atlassian.jira.rest.client.api.domain.Issue;
+import com.atlassian.jira.rest.client.api.domain.Status;
+import hudson.EnvVars;
+import hudson.Launcher;
+import hudson.model.*;
```
artkoshelev: Please avoid wildcard imports

src/test/java/hudson/plugins/jira/JiraCreateIssueNotifierTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+        when(previousBuild.getResult()).thenReturn(Result.FAILURE);
+        when(currentBuild.getResult()).thenReturn(Result.SUCCESS);
+        Assert.assertTrue(notifier.perform(currentBuild, launcher, buildListener));
+
+        // file should be deleted
+        Assert.assertEquals(0, temporaryDirectory.list().length);
+    }
+
+}
```
artkoshelev: EOL needs to be here

src/test/java/hudson/plugins/jira/JiraCreateIssueNotifierTest.java
```diff
@@ -0,0 +1,176 @@
...
+        Assert.assertTrue(notifier.perform(currentBuild, launcher, buildListener));
+
+        Assert.assertEquals(1, temporaryDirectory.list().length);
+
+        when(previousBuild.getResult()).thenReturn(Result.FAILURE);
+        when(currentBuild.getResult()).thenReturn(Result.SUCCESS);
+        Assert.assertTrue(notifier.perform(currentBuild, launcher, buildListener));
+
+        // file should be deleted
+        Assert.assertEquals(0, temporaryDirectory.list().length);
```
artkoshelev: use static imports to shorten statements

src/main/java/hudson/plugins/jira/JiraCreateIssueNotifier.java
```diff
@@ -134,25 +137,17 @@ public boolean perform(AbstractBuild<?, ?> build, Launcher launcher, BuildListen
...
         String checkDescription = (this.testDescription.equals("")) ? "No description is provided" : this.testDescription;
-        String description = String.format("The test %s has failed. \n\n%s\n\n* First failed run : [%s|%s]\n** [console log|%s]",
-                jobName, checkDescription, buildNumber, buildURL, buildURL.concat("console"));
-
-        List<Component> components = getJiraComponents(build, this.component);
-
-        String summary = "Test " + jobName + " failure - " + jenkinsURL;
+        String description = String.format("%s\n\nThe build %s has failed.\nFirst failed run: %s",
+                checkDescription, getBuildName(vars), getBuildDetailsString(vars));
+        List<BasicComponent> components = getJiraComponents(build, this.component);
```
artkoshelev: there is some code in `getJiraComponents()` converting `List<Component>` to `List<BasicComponent>` and there is same code in `session.createIssue()` converting it back :) i think you could fix it

src/test/java/hudson/plugins/jira/JiraCreateIssueNotifierTest.java
warden: There are more than 5 imports and IDEA automatically converts it to wildcard. Any reason to avoid that in tests?
Just read http://stackoverflow.com/questions/147454/why-is-using-a-wild-card-with-a-java-import-statement-bad and... there are mixed opinions..

src/main/java/hudson/plugins/jira/JiraCreateIssueNotifier.java
warden: testDescription is the preference selected in the UI, so overwriting it here will probably overwrite the plugin configuration during serialization? That's my wild guess, but I might be completely wrong.. I've assumed that also this is the reason the fields like build, job, etc.. are not class fields but local variables, because they would be treated as configuration unless transient?

src/main/java/hudson/plugins/jira/JiraCreateIssueNotifier.java
warden: This change seems to work fine.

src/main/java/hudson/plugins/jira/JiraCreateIssueNotifier.java
artkoshelev: i meant not override but have local variable with the same name

src/test/java/hudson/plugins/jira/JiraCreateIssueNotifierTest.java
artkoshelev: There is no difference between test and production code, so reasons are the same:
  * cluttering local namespace
  * with a few wildcard imports became absolutely unclear where the given types coming from

It's a matter of taste actually, but... if you don't use wildcard imports nobody will ask you to replace them with wildcard, right? :)

PR #68, opened 2015-11-17
JIRA Project Key is now expanded, so it can be given an env parameter…

tmons:
We have quite a complicated Jenkins setup, with multiple slaves and a lot of different projects. To keep things manageable, we're using the Template plugin to prevent copying the entire project configuration each time and we're relying heavily on passing parameters through environment variables.

Currently, the project key is not yet expanded by default, which for us means that we can't use the official version of this plugin at the moment. We would very much like to change that and as such, would be glad if this pr is accepted.

Kr,
Thomas Mons
In The Pocket

### DIFF
--- a/src/test/java/hudson/plugins/jira/JiraReleaseVersionUpdaterTest.java
+++ b/src/test/java/hudson/plugins/jira/JiraReleaseVersionUpdaterTest.java
@@ -23,18 +23,22 @@ import java.util.Set;
 import org.junit.Before;
 import org.junit.Test;
 import org.mockito.Mockito;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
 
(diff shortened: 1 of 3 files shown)


review-ninja:
<a href="https://app.review.ninja/jenkinsci/jira-plugin/pull/68" target="_blank">ReviewNinja

src/test/java/hudson/plugins/jira/JiraReleaseVersionUpdaterTest.java
```diff
@@ -43,45 +47,97 @@ public void createMocks() {
 		env = mock(EnvVars.class);
 		project = mock(AbstractProject.class);
 		site = mock(JiraSite.class);
+
+		when(env.expand(Mockito.anyString())).thenAnswer(new Answer<String>() {
```
warden: Can you format the lines around here ?

PR #69, opened 2015-12-17
Add JiraIssueExtractionBuilder build step to extract relevant issues …

sfhardman:
The intention of this is similar to #59 - allowing users to update Jira issues which are referenced in the commit messages for the build changeset.  I think this is a more flexible way of going about it as it can be used in conjunction with either "Progress JIRA issues by workflow action" or "Move issues matching JQL to the specified version" or custom script, and users can specify other JQL selectors as well as the issues list.

The new step extracts a comma separated list of referenced issue ID's to a properties file, which can be used in further build steps (e.g. with [EnvInject](https://wiki.jenkins-ci.org/display/JENKINS/EnvInject+Plugin))

The use case is where we are looking to update issues which are referenced in the built commits, without having to manually set the issues to a particular status in Jira to prime them for this.


review-ninja:
<a href="https://app.review.ninja/jenkinsci/jira-plugin/pull/69" target="_blank">ReviewNinja

sfhardman:
Not working right on a Windows node - on hold

PR #70, opened 2015-12-18
Split each SCM changes in paragraphs

atugushev:
Fix ugly comment message if there are *multiple changesets*. All changesets joined into one *ul* list which looks strange. They should be placed in each paragraph separately.

### Before fix
bad

### After fix
good


review-ninja:
<a href="https://app.review.ninja/jenkinsci/jira-plugin/pull/70" target="_blank">ReviewNinja

atugushev:
Thanks a lot! :+1:

PR #72, opened 2015-12-21
JENKINS-28946 - JIRA support for Workflow jobs

jan-zajic:
+ I fixed "JIRA Release Version Parameter" and "JIRA Issue Parameter" - they can be used with workflow jobs now. 
+ I repair JiraChangeLogAnnotator - it automatically work now with workflow projects. 
+ Finally i change Updater, so JJiraIssueUpdater can be used in workflow script. I used it in our build like this:
```
def scm = [$class: 'GitSCM', .... 
checkout poll: false, scm: scm
step([$class: 'hudson.plugins.jira.JiraIssueUpdater', 
    issueSelector: [$class: 'hudson.plugins.jira.DefaultUpdaterIssueSelector'], scm: scm]).
```
+ i add support for adding label to jira issue with JiraIssueUpdater, because we need it in out development workflow
+ i added option to change rest api timeout, becasuse our testing jira site is very slow ;) (JENKINS-31113)

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/jira-plugin/72)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -71,6 +71,11 @@
           <artifactId>maven-release-plugin</artifactId>
           <version>2.5</version>
         </plugin>
+        <plugin>
+          <groupId>org.jenkins-ci.tools</groupId>
+          <artifactId>maven-hpi-plugin</artifactId>
+          <version>1.95</version>
+        </plugin>
       </plugins>
     </pluginManagement>
     <plugins>
@@ -271,7 +276,18 @@
       <version>0.0.4</version>
       <scope>test</scope>
     </dependency>
-
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-scm-step</artifactId>
+      <version>1.12</version>
+      <scope>test</scope>
+    </dependency>
+   	<dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-job</artifactId>
+      <version>1.12</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
   
   <repositories>
--- a/src/main/java/hudson/plugins/jira/JiraChangeLogAnnotator.java
+++ b/src/main/java/hudson/plugins/jira/JiraChangeLogAnnotator.java
@@ -1,14 +1,5 @@
 package hudson.plugins.jira;
 
-import hudson.Extension;
-import hudson.MarkupText;
-import hudson.Util;
-import hudson.model.AbstractBuild;
-import hudson.model.AbstractProject;
-import hudson.scm.ChangeLogAnnotator;
-import hudson.scm.ChangeLogSet.Entry;
-import org.apache.commons.lang.StringUtils;
-
 import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URL;
@@ -19,6 +10,16 @@ import java.util.logging.Logger;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import org.apache.commons.lang.StringUtils;
+
+import hudson.Extension;
+import hudson.MarkupText;
+import hudson.Util;
+import hudson.model.Job;
+import hudson.model.Run;
+import hudson.scm.ChangeLogAnnotator;
+import hudson.scm.ChangeLogSet.Entry;
+
 /**
  * {@link ChangeLogAnnotator} that picks up JIRA issue IDs.
  *
@@ -26,12 +27,24 @@ import java.util.regex.Pattern;
  */
 @Extension
 public class JiraChangeLogAnnotator extends ChangeLogAnnotator {
+
     private static final Logger LOGGER = Logger.getLogger(JiraChangeLogAnnotator.class.getName());
 
+    public JiraChangeLogAnnotator() {
+        super();
+        LOGGER.fine("JiraChangeLogAnnotator created");
+    }
+
     @Override
-    public void annotate(AbstractBuild<?, ?> build, Entry change, MarkupText text) {
-        JiraSite site = getSiteForProject(build.getProject());
-        if (site == null) return;    // not configured with JIRA
+    public void annotate(Run<?, ?> build, Entry change, MarkupText text) {
+        JiraSite site = getSiteForProject(build.getParent());
+        
+        if (site == null) {
+            LOGGER.fine("not configured with JIRA");
+            return;    // not configured with JIRA
+        }
+
+        LOGGER.log(Level.FINE, "Using site: {0}", site.url);
 
         // if there's any recorded detail information, try to use that, too.
         JiraBuildAction a = build.getAction(JiraBuildAction.class);
@@ -111,7 +124,7 @@ public class JiraChangeLogAnnotator extends ChangeLogAnnotator {
         }
     }
 
-    private void saveIssues(AbstractBuild<?, ?> build, JiraBuildAction a,
+    private void saveIssues(Run<?, ?> build, JiraBuildAction a,
                             Set<JiraIssue> issuesToBeSaved) {
         if (a != null) {
             a.addIssues(issuesToBeSaved);
@@ -127,7 +140,7 @@ public class JiraChangeLogAnnotator extends ChangeLogAnnotator {
         }
     }
 
-    JiraSite getSiteForProject(AbstractProject<?, ?> project) {
+    JiraSite getSiteForProject(Job<?, ?> project) {
         return JiraSite.get(project);
     }
 }
--- a/src/main/java/hudson/plugins/jira/JiraIssueUpdater.java
+++ b/src/main/java/hudson/plugins/jira/JiraIssueUpdater.java
@@ -1,6 +1,15 @@
 package hudson.plugins.jira;
 
+import java.io.IOException;
+import java.io.PrintStream;
+import java.util.List;
+
+import org.kohsuke.stapler.DataBoundConstructor;
+
+import com.google.common.collect.Lists;
+
 import hudson.Extension;
+import hudson.FilePath;
 import hudson.Launcher;
 import hudson.matrix.MatrixAggregatable;
 import hudson.matrix.MatrixAggregator;
@@ -9,17 +18,15 @@ import hudson.matrix.MatrixRun;
 import hudson.model.AbstractBuild;
 import hudson.model.AbstractProject;
 import hudson.model.BuildListener;
+import hudson.model.Run;
+import hudson.model.TaskListener;
+import hudson.scm.SCM;
 import hudson.tasks.BuildStepDescriptor;
 import hudson.tasks.BuildStepMonitor;
 import hudson.tasks.Publisher;
 import hudson.tasks.Recorder;
 import jenkins.model.Jenkins;
-import net.sf.json.JSONObject;
-import org.kohsuke.stapler.StaplerRequest;
-
-import org.kohsuke.stapler.DataBoundConstructor;
-import java.io.IOException;
-import java.io.PrintStream;
+import jenkins.tasks.SimpleBuildStep;
 
 /**
  * Parses build changelog for JIRA issue IDs and then
@@ -27,23 +34,38 @@ import java.io.PrintStream;
  *
  * @author Kohsuke Kawaguchi
  */
-public class JiraIssueUpdater extends Recorder implements MatrixAggregatable {
+public class JiraIssueUpdater extends Recorder implements MatrixAggregatable, SimpleBuildStep {
 
     private UpdaterIssueSelector issueSelector;
+    private SCM scm;
+    private List<String> labels;
 
     @DataBoundConstructor
-    public JiraIssueUpdater(UpdaterIssueSelector issueSelector) {
+    public JiraIssueUpdater(UpdaterIssueSelector issueSelector, SCM scm, List<String> labels) {
+        super();
         this.issueSelector = issueSelector;
+        this.scm = scm;
+        if(labels != null)
+            this.labels = labels;
+        else
+            this.labels = Lists.newArrayList();
     }
 
     @Override
-    public boolean perform(AbstractBuild<?, ?> build, Launcher launcher, BuildListener listener) throws InterruptedException, IOException {
+    public void perform(Run<?, ?> run, FilePath workspace, Launcher launcher, TaskListener listener) throws InterruptedException, IOException {
         // Don't do anything for individual matrix runs.
-        if (build instanceof MatrixRun) {
-            return true;
+        if (run instanceof MatrixRun) {
+            return;
+        } else if(run instanceof AbstractBuild) {
+            AbstractBuild<?,?> abstractBuild = (AbstractBuild<?,?>) run;
+            Updater updater = new Updater(abstractBuild.getParent().getScm(), labels);
+            updater.perform(run, listener, getIssueSelector());
+        } else if(scm != null) {
+            Updater updater = new Updater(scm, labels);
+            updater.perform(run, listener, getIssueSelector());
+        } else {
+            throw new IllegalArgumentException("Unsupported run type "+run.getClass().getName());
         }
-
-        return Updater.perform(build, listener, getIssueSelector());
     }
 
     public BuildStepMonitor getRequiredMonitorService() {
@@ -70,7 +92,8 @@ public class JiraIssueUpdater extends Recorder implements MatrixAggregatable {
             public boolean endBuild() throws InterruptedException, IOException {
                 PrintStream logger = listener.getLogger();
                 logger.println("End of Matrix Build. Updating JIRA.");
-                return Updater.perform(this.build, this.listener, getIssueSelector());
+                Updater updater = new Updater(this.build.getParent().getScm(), labels);
+                return updater.perform(this.build, this.listener, getIssueSelector());
             }
         };
     }
@@ -102,4 +125,5 @@ public class JiraIssueUpdater extends Recorder implements MatrixAggregatable {
             return Jenkins.getActiveInstance().getDescriptorList(UpdaterIssueSelector.class).size() > 1;
         }
     }
+
 }
--- a/src/main/java/hudson/plugins/jira/JiraRestService.java
+++ b/src/main/java/hudson/plugins/jira/JiraRestService.java
@@ -42,13 +42,12 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
 import java.util.logging.Logger;
 
 public class JiraRestService {
 
     private static final Logger LOGGER = Logger.getLogger(JiraRestService.class.getName());
-
-    public static final int TIMEOUT_IN_10_SECONDS = 10000;
 
     public static final DateTimeFormatter DATE_TIME_FORMATTER = DateTimeFormat.forPattern("yyyy-MM-dd");
 
@@ -67,10 +66,18 @@ public class JiraRestService {
     private final String authHeader;
 
     private final String baseApiPath;
+    
+    private final int timeout;
 
+    @Deprecated
     public JiraRestService(URI uri, JiraRestClient jiraRestClient, String username, String password) {
+    	this(uri, jiraRestClient, username, password, JiraSite.DEFAULT_TIMEOUT);
+    }
+
+    public JiraRestService(URI uri, JiraRestClient jiraRestClient, String username, String password, int timeout) {
         this.uri = uri;
         this.objectMapper = new ObjectMapper();
+        this.timeout = timeout;
         final String login = username + ":" + password;
         try {
             byte[] encodeBase64 = Base64.encodeBase64(login.getBytes("UTF-8"));
@@ -109,15 +116,15 @@ public class JiraRestService {
         }
 
       try {
-          jiraRestClient.getIssueClient().addComment(builder.build(), comment).get(10, TimeUnit.SECONDS);
+          jiraRestClient.getIssueClient().addComment(builder.build(), comment).get(timeout, TimeUnit.SECONDS);
       } catch (Exception e) {
           LOGGER.warning("jira rest client add comment error. cause: " + e.getMessage());
       }
     }
-
+  
     public Issue getIssue(String issueKey) {
         try {
-            return jiraRestClient.getIssueClient().getIssue(issueKey).get(10, TimeUnit.SECONDS);
+            return jiraRestClient.getIssueClient().getIssue(issueKey).get(timeout, TimeUnit.SECONDS);
         } catch (Exception e) {
             LOGGER.warning("jira rest client get issue error. cause: " + e.getMessage());
             return null;
@@ -126,7 +133,7 @@ public class JiraRestService {
 
     public List<IssueType> getIssueTypes() {
         try {
-            return Lists.newArrayList(jiraRestClient.getMetadataClient().getIssueTypes().get(10, TimeUnit.SECONDS));
+            return Lists.newArrayList(jiraRestClient.getMetadataClient().getIssueTypes().get(timeout, TimeUnit.SECONDS));
         } catch (Exception e) {
             LOGGER.warning("jira rest client get issue types error. cause: " + e.getMessage());
             return Collections.emptyList();
@@ -136,7 +143,7 @@ public class JiraRestService {
     public List<String> getProjectsKeys() {
         Iterable<BasicProject> projects = Collections.emptyList();
         try {
-            projects = jiraRestClient.getProjectClient().getAllProjects().get(10, TimeUnit.SECONDS);
+            projects = jiraRestClient.getProjectClient().getAllProjects().get(timeout, TimeUnit.SECONDS);
         } catch (Exception e) {
             LOGGER.warning("jira rest client get project keys error. cause: " + e.getMessage());
         }
@@ -151,7 +158,7 @@ public class JiraRestService {
         try {
             final SearchResult searchResult = jiraRestClient.getSearchClient()
                                                             .searchJql(jqlSearch, maxResults, 0, null)
-                                                            .get(10, TimeUnit.SECONDS);
+                                                            .get(timeout, TimeUnit.SECONDS);
             return Lists.newArrayList(searchResult.getIssues());
         } catch (Exception e) {
             LOGGER.warning("jira rest client get issue from jql search error. cause: " + e.getMessage());
@@ -192,7 +199,7 @@ public class JiraRestService {
         final VersionInput versionInput = new VersionInput(projectKey, versionName, null, null, false, false);
         try {
             return jiraRestClient.getVersionRestClient()
-                          .createVersion(versionInput).get(10, TimeUnit.SECONDS);
+                          .createVersion(versionInput).get(timeout, TimeUnit.SECONDS);
         } catch (Exception e) {
             LOGGER.warning("jira rest client add version error. cause: " + e.getMessage());
             return null;
@@ -207,7 +214,7 @@ public class JiraRestService {
             .getReleaseDate(), version.isArchived(), version.isReleased());
 
         try {
-            jiraRestClient.getVersionRestClient().updateVersion(builder.build(), versionInput).get(10, TimeUnit.SECONDS);
+            jiraRestClient.getVersionRestClient().updateVersion(builder.build(), versionInput).get(timeout, TimeUnit.SECONDS);
         }catch (Exception e) {
             LOGGER.warning("jira rest client release version error. cause: " + e.getMessage());
         }
@@ -229,7 +236,7 @@ public class JiraRestService {
         final IssueInput issueInput = builder.build();
 
         try {
-            return jiraRestClient.getIssueClient().createIssue(issueInput).get(10, TimeUnit.SECONDS);
+            return jiraRestClient.getIssueClient().createIssue(issueInput).get(timeout, TimeUnit.SECONDS);
         } catch (Exception e) {
             LOGGER.warning("JIRA REST createIssue error: " + e.getMessage());
             return null;
@@ -238,30 +245,41 @@ public class JiraRestService {
 
     public User getUser(String username) {
         try {
-            return jiraRestClient.getUserClient().getUser(username).get(10, TimeUnit.SECONDS);
+            return jiraRestClient.getUserClient().getUser(username).get(timeout, TimeUnit.SECONDS);
         } catch (Exception e) {
             LOGGER.warning("jira rest client get user error. cause: " + e.getMessage());
             return null;
         }
     }
-
+ 
     public void updateIssue(String issueKey, List<Version> fixVersions) {
         final IssueInput issueInput = new IssueInputBuilder().setFixVersions(fixVersions)
                                                              .build();
         try {
-            jiraRestClient.getIssueClient().updateIssue(issueKey, issueInput).get(10, TimeUnit.SECONDS);
+            jiraRestClient.getIssueClient().updateIssue(issueKey, issueInput).get(timeout, TimeUnit.SECONDS);
         } catch (Exception e) {
             LOGGER.warning("jira rest client update issue error. cause: " + e.getMessage());
         }
     }
-
+    
+    public void setIssueLabels(String issueKey, List<String> labels) {
+        final IssueInput issueInput = new IssueInputBuilder()
+        		.setFieldValue(IssueFieldId.LABELS_FIELD.id, labels)
+                .build();
+        try {
+            jiraRestClient.getIssueClient().updateIssue(issueKey, issueInput).get(timeout, TimeUnit.SECONDS);
+        } catch (Exception e) {
+            LOGGER.log(Level.WARNING, "jira rest client update labels error for issue "+issueKey, e);
+        }
+    }    
+    
     public Issue progressWorkflowAction(String issueKey, Integer actionId) {
         final TransitionInput transitionInput = new TransitionInput(actionId);
 
         final Issue issue = getIssue(issueKey);
 
         try {
-            jiraRestClient.getIssueClient().transition(issue, transitionInput).get(10, TimeUnit.SECONDS);
+            jiraRestClient.getIssueClient().transition(issue, transitionInput).get(timeout, TimeUnit.SECONDS);
         } catch (Exception e) {
             LOGGER.warning("jira rest client process workflow action error. cause: " + e.getMessage());
         }
@@ -274,7 +292,7 @@ public class JiraRestService {
         try {
             final Iterable<Transition> transitions = jiraRestClient.getIssueClient()
                                                                    .getTransitions(issue)
-                                                                   .get(10, TimeUnit.SECONDS);
+                                                                   .get(timeout, TimeUnit.SECONDS);
             return Lists.newArrayList(transitions);
         } catch (Exception e) {
             LOGGER.warning("jira rest client get available actions error. cause: " + e.getMessage());
@@ -285,7 +303,7 @@ public class JiraRestService {
     public List<Status> getStatuses() {
         try {
             final Iterable<Status> statuses = jiraRestClient.getMetadataClient().getStatuses()
-                                                            .get(10, TimeUnit.SECONDS);
+                                                            .get(timeout, TimeUnit.SECONDS);
             return Lists.newArrayList(statuses);
         } catch (Exception e) {
             LOGGER.warning("jira rest client get statuses error. cause: " + e.getMessage());
@@ -329,11 +347,15 @@ public class JiraRestService {
 
     private Request buildGetRequest(URI uri) {
         return Request.Get(uri)
-                .connectTimeout(TIMEOUT_IN_10_SECONDS)
-                .socketTimeout(TIMEOUT_IN_10_SECONDS)
+                .connectTimeout(timeoutInMiliseconds())
+                .socketTimeout(timeoutInMiliseconds())
                 .addHeader("Authorization", authHeader)
                 .addHeader("Content-Type", "application/json");
     }
+
+	protected int timeoutInMiliseconds() {
+		return (int) TimeUnit.SECONDS.toMillis(timeout);
+	}
 
     public String getBaseApiPath() {
         return baseApiPath;
--- a/src/main/java/hudson/plugins/jira/JiraSession.java
+++ b/src/main/java/hudson/plugins/jira/JiraSession.java
@@ -1,7 +1,6 @@
 package hudson.plugins.jira;
 
-import com.atlassian.jira.rest.client.api.domain.*;
-import com.google.common.collect.Lists;
+import static org.apache.commons.lang.StringUtils.isNotEmpty;
 
 import java.util.HashMap;
 import java.util.HashSet;
@@ -9,7 +8,15 @@ import java.util.List;
 import java.util.Set;
 import java.util.logging.Logger;
 
-import static org.apache.commons.lang.StringUtils.isNotEmpty;
+import com.atlassian.jira.rest.client.api.domain.BasicIssue;
+import com.atlassian.jira.rest.client.api.domain.Component;
+import com.atlassian.jira.rest.client.api.domain.Issue;
+import com.atlassian.jira.rest.client.api.domain.IssueType;
+import com.atlassian.jira.rest.client.api.domain.Permissions;
+import com.atlassian.jira.rest.client.api.domain.Status;
+import com.atlassian.jira.rest.client.api.domain.Transition;
+import com.atlassian.jira.rest.client.api.domain.Version;
+import com.google.common.collect.Lists;
 
 /**
  * Connection to JIRA.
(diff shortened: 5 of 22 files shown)


review-ninja:
<a href="https://app.review.ninja/jenkinsci/jira-plugin/pull/72" target="_blank">ReviewNinja

warden:
This is a great piece of work, I will definitely take a look into it further to test it out this week, but it looks great. One thing that comes to my mind, the dependency for workflow plugin - maybe it can be optional, for  people who are not using it currently?

jan-zajic:
I use a class constant for the default 10 seconds timeout, as you wished.

jan-zajic:
I mark dependency on workflow-job as optional, dep on workflow-scm-step has only test scope. Can someone test original functionality on some jenkins installation without wf plugin? I hope that all original stuff works and i do some basic test, but i can't proof it.

warden:
Thanks @jan-zajic for all the fixes. I will try to merge this soon, when I will review the Workflow changes.:+1: good job! 
In the future, please try to open separate PRs for separate issues as this speeds up merging process (i.e. with the current PR, I would gladly merge Timeout and Labels commits, but I need to take a closer look on the Workflow changes). Also, if you can, new tests are welcome where possible for uncovered parts of code :)

jan-zajic:
ok, thanks You for review and merging it. I can post aditional pull requests, for other issues, after this main change is merged.

src/main/java/hudson/plugins/jira/Updater.java
```diff
@@ -348,10 +406,29 @@ static void findIssues(AbstractBuild<?, ?> build, Set<String> ids, Pattern patte
...
-    /**
-     * Debug flag.
-     */
-    public static boolean debug = false;
+    private List<ChangeLogSet<? extends Entry>> getChanges(Run<?,?> run)
+    {
+        if(run instanceof AbstractBuild)
+        	return ((AbstractBuild<?,?>) run).getChangeSets();
+        else if(run instanceof WorkflowRun)
+        	return ((WorkflowRun)run).getChangeSets();
```
jglick: Pending this aspect of [JENKINS-24141](https://issues.jenkins-ci.org/browse/JENKINS-24141), I recommend looking for a `getChangeSets` method via reflection, rather than adding (even an `optional`) plugin dependency on `workflow-job`.

jan-zajic: ok, it's ugly but i can do it, if it's preffered. what do you think, @warden ?

warden:
I hoped that someone well versed with Jenkins API will take a look at this, thanks @jglick ! 
I agree with @jglick, as this will avoid the optional dependency and possible NPEs and compatibility issues.

warden:
I hoped that someone well versed with Jenkins API will take a look at this, thanks @jglick ! 
I agree with @jglick, as this will avoid the optional dependency and possible NPEs and compatibility issues.
@jan-zajic if you need more time to work on this and want other changes to be merged (timeout, label assignment), please split them into separate into separate PR.

jan-zajic:
I'm done with it, see Updater.getChangesUsingReflection method and coresponding unit tests.

src/main/java/hudson/plugins/jira/JiraBuildAction.java
```diff
@@ -11,11 +15,12 @@
  * @author Kohsuke Kawaguchi
  */
 public class JiraBuildAction implements Action {
-    public final AbstractBuild<?, ?> owner;
+    
+	public final Run<?, ?> owner;
```
oleg-nenashev: Binary compatibility break BTW. Requires a major version bump + review of dependent plugins

src/main/java/hudson/plugins/jira/JiraBuildAction.java
```diff
@@ -11,11 +15,12 @@
...
 
     public JiraIssue[] issues;
 
-    public JiraBuildAction(AbstractBuild<?, ?> owner, Collection<JiraIssue> issues) {
+    public JiraBuildAction(Run<?, ?> owner, Collection<JiraIssue> issues) {
```
oleg-nenashev: doh

src/main/java/hudson/plugins/jira/JiraChangeLogAnnotator.java
```diff
@@ -26,13 +27,36 @@
...
 public class JiraChangeLogAnnotator extends ChangeLogAnnotator {
+	
     private static final Logger LOGGER = Logger.getLogger(JiraChangeLogAnnotator.class.getName());
 
+    public JiraChangeLogAnnotator()
+    {
+        super();
+        if(LOGGER.isLoggable(Level.FINE))
+        {
+        	LOGGER.info("JiraChangeLogAnnotator registered");
```
oleg-nenashev: LOGGER.log(Level.FINE, "JiraChangeLogAnnotator registered")
BTW it does not guaranteed that annotator is registered. You would have to check Jenkins' extension lists

src/main/java/hudson/plugins/jira/JiraChangeLogAnnotator.java
```diff
@@ -26,13 +27,36 @@
...
+        	if(LOGGER.isLoggable(Level.FINE))
+            {
+        		LOGGER.fine("not configured with JIRA");
+            }
+        	return;    // not configured with JIRA
+    	}
+        
+        if(LOGGER.isLoggable(Level.FINE))
+        {
+        	LOGGER.fine("Using site: " + site.url);
```
oleg-nenashev: LOGGER.log(Level.FINE, "Using site: {0}", site.url)

src/main/java/hudson/plugins/jira/JiraChangeLogAnnotator.java
```diff
@@ -127,7 +151,7 @@ private void saveIssues(AbstractBuild<?, ?> build, JiraBuildAction a,
         }
     }
 
-    JiraSite getSiteForProject(AbstractProject<?, ?> project) {
+    JiraSite getSiteForProject(Job<?, ?> project) {
```
oleg-nenashev: If you break the binary compatibility, consider using an <code>Item</code>

src/main/java/hudson/plugins/jira/JiraIssueUpdater.java
```diff
@@ -27,25 +33,39 @@
...
+	}
+	
+	@Override
+	public void perform(Run<?, ?> run, FilePath workspace, Launcher launcher, TaskListener listener) throws InterruptedException, IOException 
+	{
+		if(run instanceof AbstractBuild)
+		{
+			AbstractBuild<?,?> abstractBuild = (AbstractBuild<?,?>) run;
+			Updater updater = new Updater(abstractBuild.getParent().getScm(), labels);
+			updater.perform(run, listener, getIssueSelector());
```
oleg-nenashev: Breaks the behavior for <code> MatrixRun</code>s as far as I see

src/main/java/hudson/plugins/jira/JiraRestService.java
```diff
@@ -48,7 +48,7 @@
 
     private static final Logger LOGGER = Logger.getLogger(JiraRestService.class.getName());
 
-    public static final int TIMEOUT_IN_10_SECONDS = 10000;
+    public static final int MILIS_IN_ONE_SECOND = 1000;
```
oleg-nenashev: Avoid public magic values. Now it breaks the binary compatibility

src/main/java/hudson/plugins/jira/JiraRestService.java
```diff
@@ -67,10 +67,13 @@
     private final String authHeader;
 
     private final String baseApiPath;
+    
+    private final int timeout;
 
-    public JiraRestService(URI uri, JiraRestClient jiraRestClient, String username, String password) {
+    public JiraRestService(URI uri, JiraRestClient jiraRestClient, String username, String password, int timeout) {
```
oleg-nenashev: To retain the binary compatibility you would need to keep the original constructor and to mark it as deprecated

src/main/java/hudson/plugins/jira/JiraRestService.java
```diff
@@ -238,22 +241,33 @@ public BasicIssue createIssue(String projectKey, String description, String assi
...
     }
+    
+    public void setIssueLabels(String issueKey, List<String> labels) {
+        final IssueInput issueInput = new IssueInputBuilder()
+        		.setFieldValue(IssueFieldId.LABELS_FIELD.id, labels)
+                .build();
+        try {
+            jiraRestClient.getIssueClient().updateIssue(issueKey, issueInput).get(timeout, TimeUnit.SECONDS);
+        } catch (Exception e) {
+            LOGGER.warning("jira rest client update issue error. cause: " + e.getMessage());
```
oleg-nenashev: Also keep the stacktrace in the log

src/main/java/hudson/plugins/jira/JiraRestService.java
oleg-nenashev: I would not expect it ever changes (proof?), so it's OK to keep directly in the method

lanwen: use `TimeUnit.SECONDS.toMillis(long)` to convert time values

pom.xml
```diff
@@ -71,6 +71,11 @@
           <artifactId>maven-release-plugin</artifactId>
           <version>2.5</version>
         </plugin>
+		<plugin>
```
KostyaSha: wrong indent

src/main/java/hudson/plugins/jira/JiraChangeLogAnnotator.java
jan-zajic: It's not important, i only log it with fine level and use it when debuging plugin.. Ok, i change message to ""JiraChangeLogAnnotator created".

src/main/java/hudson/plugins/jira/JiraChangeLogAnnotator.java
jan-zajic: Ok i change it, it's maybe nicer, or what's the point??

lanwen: logger should check by itself the level in this form, so no need to `if(LOGGER.isLoggable(Level.FINE))`

src/main/java/hudson/plugins/jira/JiraRestService.java
jan-zajic: I will change it to TimeUnit.SECONDS.toMillis as @lanwen point, thank you for good advice!

src/main/java/hudson/plugins/jira/JiraChangeLogAnnotator.java
```diff
@@ -26,13 +27,36 @@
...
+    
     @Override
-    public void annotate(AbstractBuild<?, ?> build, Entry change, MarkupText text) {
-        JiraSite site = getSiteForProject(build.getProject());
-        if (site == null) return;    // not configured with JIRA
-
+    public void annotate(Run<?, ?> build, Entry change, MarkupText text) {
+        JiraSite site = getSiteForProject(build.getParent());
+        
+        if (site == null)
```
lanwen: please use oracle style formatting with brackets on the same line

src/main/java/hudson/plugins/jira/JiraRestService.java
jan-zajic: OK, i will do this.

src/main/java/hudson/plugins/jira/JiraRestService.java
jan-zajic: ok, i will change it, but it will be different than rest of the class, maybe better to create separate PR in future ti fix all logging lines?

pom.xml
jan-zajic: ok, i fix it.

src/main/java/hudson/plugins/jira/JiraIssueUpdater.java
jan-zajic: ?? how ??

src/main/java/hudson/plugins/jira/JiraChangeLogAnnotator.java
jan-zajic: This method i currently used only in same class and with job, i can change it in future if some relevant issue exists.

src/main/java/hudson/plugins/jira/JiraRestService.java
lanwen: new lines should be added with the most relevant rules, as of new PR can be merged not so soon :)

src/main/java/hudson/plugins/jira/JiraSite.java
```diff
@@ -148,7 +161,7 @@
 
     @DataBoundConstructor
     public JiraSite(URL url, URL alternativeUrl, String userName, String password, boolean supportsWikiStyleComment, boolean recordScmChanges, String userPattern,
-                    boolean updateJiraIssueForAllStatus, String groupVisibility, String roleVisibility, boolean useHTTPAuth) {
+                    boolean updateJiraIssueForAllStatus, String groupVisibility, String roleVisibility, boolean useHTTPAuth, Integer timeout) {
```
KostyaSha: @lanwen start using DBSetters?

src/main/java/hudson/plugins/jira/JiraSite.java
```diff
@@ -164,12 +177,20 @@ public JiraSite(URL url, URL alternativeUrl, String userName, String password, b
             }
 
         this.url = url;
+        
+        if(timeout != null) {
+        	this.timeout = timeout;
+        } else {
+        	this.timeout = JiraSite.DEFAULT_TIMEOUT;
```
KostyaSha: seems @lanwen should know one liner with default....

src/main/java/hudson/plugins/jira/JiraSite.java
lanwen: there is more robust
`@DataBoundSetter`
https://groups.google.com/forum/#!msg/jenkinsci-dev/58-DEvuJZWI/5QrxBZRFJ6IJ
http://stapler.kohsuke.org/apidocs/org/kohsuke/stapler/DataBoundSetter.html

src/main/java/hudson/plugins/jira/JiraChangeLogAnnotator.java
oleg-nenashev: See the comment from @lanwen below. These loggers do not require prechecks

src/main/java/hudson/plugins/jira/JiraIssueUpdater.java
oleg-nenashev: Original behavior: 

```java
// Don't do anything for individual matrix runs.
         if (build instanceof MatrixRun) {
             return true;
        }
```

Now the code handle matrix runs as other abstract builds

src/main/java/hudson/plugins/jira/JiraSite.java
oleg-nenashev: E.g. https://commons.apache.org/proper/commons-lang/javadocs/api-3.1/org/apache/commons/lang3/ObjectUtils.html#defaultIfNull(T,%20T)

lanwen: if u use dbsetter, no need to precheck for null. Simply use default value in constructor (and make field not final).

src/main/java/hudson/plugins/jira/JiraBuildAction.java
jan-zajic: I don't think so. Run is super class of AbstractBuild so no java compatibility can be broken. If you think compatibility with serialization of jobs and runs, i read in hudson.model.Action: Actions are often persisted as a part of {@link Actionable} (for example with {@link Build}) via XStream. If actions are persited via XStream, it's not problem as i known XStream. I add junit test in next commit to this PR that proof it.

src/main/java/hudson/plugins/jira/JiraBuildAction.java
jan-zajic: see my comment above

src/main/java/hudson/plugins/jira/JiraSite.java
KostyaSha: Better in getter. Null will be in field after xstream deserialisation.

src/main/java/hudson/plugins/jira/Updater.java
```diff
@@ -72,10 +110,11 @@ static boolean perform(AbstractBuild<?, ?> build, BuildListener listener, Update
             }
 
             boolean doUpdate = false;
-            if (site.updateJiraIssueForAllStatus) {
+            //in case of workflow, it may be null
+            if (site.updateJiraIssueForAllStatus || build.getResult() == null) {
                 doUpdate = true;
             } else {
-                doUpdate = build.getResult().isBetterOrEqualTo(Result.UNSTABLE);
+                doUpdate =  build.getResult().isBetterOrEqualTo(Result.UNSTABLE);
```
KostyaSha: remove space

src/main/java/hudson/plugins/jira/Updater.java
```diff
@@ -124,13 +164,15 @@ static void submitComments(
...
             logger.println(Messages.Updater_Updating(issue.id));
-
+            
             try {
                 session.addComment(
                         issue.id,
                         createComment(build, useWikiStyleComments, jenkinsRootUrl, recordScmChanges, issue),
                         groupVisibility, roleVisibility
                 );
+                if(!labels.isEmpty())
```
KostyaSha: brackets for if, space after if

src/main/java/hudson/plugins/jira/Updater.java
```diff
@@ -124,13 +164,15 @@ static void submitComments(
 
         for (JiraIssue issue : copy) {
             logger.println(Messages.Updater_Updating(issue.id));
-
+
```
KostyaSha: remove spaces

src/main/java/hudson/plugins/jira/JiraChangeLogAnnotator.java
jan-zajic: Ok i will remove checks

src/main/java/hudson/plugins/jira/JiraBuildAction.java
oleg-nenashev: Binary compatibility also includes API.
If somebody uses <code>AbstractBuils<?,?> build = x.getAction(JiraBuildAction.class)owner</code> somewhere in the dependent code, this method will become non-compilable and cause exceptions in Runtume

src/main/java/hudson/plugins/jira/JiraIssueUpdater.java
jan-zajic: Thank you, i will fix it.

src/main/java/hudson/plugins/jira/JiraBuildAction.java
oleg-nenashev: Regarding the serialization, the right way in new Jenkins API is to implement RunAction2 interface andmake the field transient

src/main/java/hudson/plugins/jira/JiraRestService.java
jan-zajic: OK :+1:

src/main/java/hudson/plugins/jira/JiraChangeLogAnnotator.java
jan-zajic: OK

src/main/java/hudson/plugins/jira/JiraSite.java
jan-zajic: ok, i change it to DataBoundSetter, it's a good idea

src/main/java/hudson/plugins/jira/Updater.java
jan-zajic: ok

src/main/java/hudson/plugins/jira/Updater.java
jan-zajic: ok

src/main/java/hudson/plugins/jira/Updater.java
jan-zajic: ok

src/main/java/hudson/plugins/jira/JiraBuildAction.java
jan-zajic: I know of no plugin that currently depends on jira plugin. Is anywhere documentation of this dependencies or some other way to find it? If someone have plugin that depends on jira plugin, please point me and i check if this change is problem.

src/main/java/hudson/plugins/jira/JiraSession.java
```diff
@@ -66,6 +73,25 @@ public void addComment(String issueId, String comment,
...
+    	boolean changed = false;
+    	for(String label : labels)
+    	{
+    		if(!newLabels.contains(label))
+    		{    		
+    			newLabels.add(label);
+    			changed = true;
+    		}
+    	}
+    	if(changed)
```
lanwen: please add brackets

src/main/java/hudson/plugins/jira/JiraSession.java
```diff
@@ -66,6 +73,25 @@ public void addComment(String issueId, String comment,
...
 
+    public void addLabels(String issueId, List<String> labels)
+    {
+    	List<String> newLabels = Lists.newArrayList();
+    	Issue existingIssue = service.getIssue(issueId);
+    	if(existingIssue.getLabels() != null)
+    		newLabels.addAll(existingIssue.getLabels());
+    	boolean changed = false;
+    	for(String label : labels)
+    	{
```
lanwen: move this to prev line (oracle style)

src/main/java/hudson/plugins/jira/JiraChangeLogAnnotator.java
```diff
@@ -26,13 +27,26 @@
  */
 @Extension
 public class JiraChangeLogAnnotator extends ChangeLogAnnotator {
+	
     private static final Logger LOGGER = Logger.getLogger(JiraChangeLogAnnotator.class.getName());
 
+    public JiraChangeLogAnnotator()
```
lanwen: sorry for bothering, but please change same style for brackets everywhere to oracle style :)

src/test/java/hudson/plugins/jira/JiraChangeLogAnnotatorTest.java
```diff
@@ -65,6 +75,22 @@ public void testAnnotate() throws Exception {
...
+        when(b.getAction(JiraBuildAction.class)).thenReturn(new JiraBuildAction(b, Collections.singleton(new JiraIssue("DUMMY-1", TITLE))));
+
+        MarkupText text = new MarkupText("marking up DUMMY-1.");
+        JiraChangeLogAnnotator annotator = spy(new JiraChangeLogAnnotator());
+        doReturn(site).when(annotator).getSiteForProject((AbstractProject<?, ?>) Mockito.any());
+
+        annotator.annotate(b, null, text);
+
+        // make sure '$' didn't confuse the JiraChangeLogAnnotator
+        Assert.assertTrue(text.toString(false).contains(TITLE));
```
lanwen: please use form `assertThat(text.toSting(false), containsString(TITLE));` as more verbosibe in case of fail

lanwen:
Please pay some attention to code style, otherwise any other contributor will want to change formatting and make dummy commits, or worse than that - every contributor will change style according to his opinion and this can produce very hard to read code.

jan-zajic:
I have nothing against code formatting, but it's difficult to be comply with standard, if it does not exists. Where is written which code style this project use? And what about to add maven-java-formatter-plugin or some other build check to project to ensure right formatting? If you preffer Oracle Code Conventions for the Java, please write it on github title page of project, but it is very old and not maintained (if we think same document) so preffer to publish some file with preffered format (eclipse, checkstyle, whatever.. ).

oleg-nenashev:
@jan-zajic 
In Jenkins plugins plugin owners are eligible to define their own code styles.
So better to ask @warden about his requirements.

In Jenkins core we advertise Oracle code style, but de-facto it's not being strictly followed. IMHO not a big issue while code is easily readable

lanwen:
Great part of this plugin uses oracle style (with some exceptions). But your PR contains mixed style of all existing formatting rules and indentation :) so if you can, you should use autoformatter of your IDE to avoid such mix.

Sorry for intent observation of this problem, but modern IDE can format code in a pretty way for most of people with only a few clicks :) Use this power for new code and all will be ok

src/test/java/hudson/plugins/jira/JiraChangeLogAnnotatorTest.java
```diff
@@ -65,6 +73,22 @@ public void testAnnotate() throws Exception {
         Assert.assertTrue(text.toString(false).contains(TITLE));
     }
 
+    @Test
+    public void testAnnotateWf() throws Exception {
+    	Run b = mock(Run.class);
```
warden: Needs indent fix on the next lines

src/test/java/hudson/plugins/jira/UpdaterTest.java
```diff
@@ -31,6 +61,8 @@
 @SuppressWarnings("unchecked")
 public class UpdaterTest {
 
+	private Updater updater;
+
```
warden: Both lines, wrong indent.

src/test/java/hudson/plugins/jira/deprecated/DeprecatedJiraBuildAction.java
```diff
@@ -0,0 +1,57 @@
...
+
+import hudson.model.AbstractBuild;
+import hudson.model.Action;
+import hudson.plugins.jira.JiraIssue;
+import hudson.plugins.jira.Messages;
+
+import java.util.*;
+
+/**
+ * JIRA issues related to the build.
```
warden: Needs a description on why this class is here and what do the corresponding tests test in fact.
Also, please change or remove @author.

pom.xml
```diff
@@ -71,6 +71,11 @@
           <artifactId>maven-release-plugin</artifactId>
           <version>2.5</version>
         </plugin>
+        <plugin>
+          <groupId>org.jenkins-ci.tools</groupId>
+          <artifactId>maven-hpi-plugin</artifactId>
+          <version>1.95</version>
```
lanwen: why exactly this version?

src/main/java/hudson/plugins/jira/JiraBuildAction.java
warden: @oleg-nenashev as I understand you mean other Jenkins plugins that depend on this one, are you aware of any? (just by chance). So, we should keep the old API with @deprecated for one release along with introducing the new API and that would be sufficient, right?
Also, I'm curious how do you know about RunAction2, that's good to know, but I don't see any description in the [API doc for this](http://javadoc.jenkins-ci.org/jenkins/model/RunAction2.html). Is this just your Jenkins core source knowledge from exp? :)

src/main/java/hudson/plugins/jira/JiraChangeLogAnnotator.java
warden: I agree that using Item is not part of this change, but still good to learn about this, thanks :+1:

warden:
Wow, this discussion really exploded. Thanks all for input and reviews!
I can see that most of the issues were now addressed, some minor things left.

As far as the formatting is concerned, the standard for me is to mirror the same formatting rules that were used before in the code changed. So new code should be added with the same formatting.
Exceptions like tabs to space conversion for the whole file should be done as a separate commit to preserve change visibility.

When all comments are resolved, please squash relevant commits - this is to preserve the commit log from being polluted of irrelevant commit messages (like "oracle style formatting" of the actually changed code).

Thanks, @jan-zajic good job, I really look forward to get this merged.

pom.xml
warden: Probably because this is the current one :)
https://wiki.jenkins-ci.org/display/JENKINS/Workflow+Plugin

warden: True, better to use the current one: 1.115

src/main/java/hudson/plugins/jira/JiraChangeLogAnnotator.java
warden: !completed

src/main/java/hudson/plugins/jira/JiraSession.java
warden: !fixed

src/main/java/hudson/plugins/jira/JiraSession.java
warden: !fixed

pom.xml
warden: !fixed

src/main/java/hudson/plugins/jira/JiraChangeLogAnnotator.java
warden: !fixed

src/main/java/hudson/plugins/jira/JiraChangeLogAnnotator.java
warden: !fixed

src/main/java/hudson/plugins/jira/JiraChangeLogAnnotator.java
warden: !fixed

src/main/java/hudson/plugins/jira/JiraIssueUpdater.java
warden: !fixed

jan-zajic:
@warden, version 1.115 of maven-hpi-plugin add some new requirements to jelly files through junit tests, see last build result https://jenkins.ci.cloudbees.com/job/plugins/job/jira-plugin/181/ after upgrading it https://github.com/jan-zajic/jira-plugin/commit/780eefbcb0dae68b036377f8fd9ab91b68bc50ac.  Should i fix this issues, or revert version to 1.95?

warden:
As you are adding this dependency, it would be actually welcome to fix the issues around it while adding the newest version possible.  OTOH, I'm fine with going with an older version while not touching unrelated view files.
So in this case, whatever you prefer actually :)

jan-zajic:
I leave 1.95. I originally want to upgrade it least as possible, because this PR add many unrelated changes even without this. 
I can upgrade to current version in separate PR later, if you want, create ticket in jira and assign it to me.

src/main/java/hudson/plugins/jira/UpdaterIssueSelector.java
```diff
@@ -27,6 +27,6 @@
      * @param listener Current's run listener.
      * @return Set of ids of issues which should be updated.
      */
-    public abstract Set<String> findIssueIds(@Nonnull Run<?, ?> run, @Nonnull JiraSite site, @Nonnull TaskListener listener);
+    public abstract Set<String> findIssueIds(@Nonnull Run<?, ?> run, @Nonnull Updater updater, @Nonnull JiraSite site, @Nonnull TaskListener listener);
```
FrantaM: API break. Btw should `Updater` really be public?

warden: Second time somebody mentions API break. 
@jan-zajic can you preserve the old API with @deprecated?

src/main/java/hudson/plugins/jira/JiraBuildAction.java
oleg-nenashev: @warden @jan-zajic
I'm not aware about plugins utilizing this API as well. 
On the other hand I've definitely used this action within [Scriptler](https://wiki.jenkins-ci.org/display/JENKINS/Scriptler+Plugin) scripts at my previous work. Such valid usages of public API are also under risk. Not sure about my instance, I do not recall exact types and methods unfortunately.

> Also, I'm curious how do you know about RunAction2, that's good to know, but I don't see any description in the API doc for this. Is this just your Jenkins core source knowledge from exp? :)

Yep, just a sacred knowledge I've got during Workflow integration of other plugin. Documentation needs improvement, of course

src/main/java/hudson/plugins/jira/UpdaterIssueSelector.java
jan-zajic: I prefer not change this public method, but default implementation used Updater way that requires refactoring if we want to keep this api unchanged. I try to think about it.

jan-zajic: I overlooked, it's not need at all, i will fix it.

jan-zajic:
@FrantaM., @warden I removed unnecessary public api break. I'm glad we are getting to the important issues.

jan-zajic:
@warden I do not understand what's the problem with building PR after last commit. It build and test localy just find. Can you, or someone else, check it, please ?

jan-zajic:
@warden  - should i squash relevant commits together now, as you mentoined? Maybe it will help with build issue.

warden:
I will fix the build issue later today. Yes, please squash them, you should
have around three commits here for those changes .

warden:
The build should be now fixed, please rebase to fix conflicts.

jan-zajic:
ok, I'm done with it, It was quite hard.

warden:
@oleg-nenashev , @lanwen any comments on this in the current shape?

src/main/java/hudson/plugins/jira/Updater.java
```diff
@@ -359,10 +417,62 @@ static void findIssues(AbstractBuild<?, ?> build, Set<String> ids, Pattern patte
...
+                throw new IllegalStateException(CANNOT_ACCESS_GET_CHANGESET_METHOD, e);
+            }
+        } else {
+            //if run don't have  GET_CHANGESET_METHOD, we don't support it
+            throw new IllegalArgumentException("Unsupported Run type "+run.getClass().getName());
+        }
+    }
+
+    private static Map<AbstractProject,DependencyChange> getDependencyChanges(Run<?,?> run) {
+        AbstractBuild<?,?> build = null;
```
warden: Unused

src/main/java/hudson/plugins/jira/JiraChangeLogAnnotator.java
```diff
@@ -26,12 +27,24 @@
  */
 @Extension
 public class JiraChangeLogAnnotator extends ChangeLogAnnotator {
+
     private static final Logger LOGGER = Logger.getLogger(JiraChangeLogAnnotator.class.getName());
 
+    public JiraChangeLogAnnotator() {
+        super();
+    	LOGGER.info("JiraChangeLogAnnotator created");
```
lanwen: Think it should be `fine` lvl

src/main/java/hudson/plugins/jira/JiraIssueUpdater.java
```diff
@@ -27,23 +34,38 @@
...
+            return;
+        } else if(run instanceof AbstractBuild) {
+            AbstractBuild<?,?> abstractBuild = (AbstractBuild<?,?>) run;
+            Updater updater = new Updater(abstractBuild.getParent().getScm(), labels);
+            updater.perform(run, listener, getIssueSelector());
+        } else if(scm != null) {
+            Updater updater = new Updater(scm, labels);
+            updater.perform(run, listener, getIssueSelector());
+        } else {
+            throw new IllegalArgumentException("Unsupported job type "+run.getClass().getName());
```
lanwen: Run is not a job

src/main/java/hudson/plugins/jira/JiraRestService.java
```diff
@@ -238,30 +245,41 @@ public BasicIssue createIssue(String projectKey, String description, String assi
...
-
+    
+    public void setIssueLabels(String issueKey, List<String> labels) {
+        final IssueInput issueInput = new IssueInputBuilder()
+        		.setFieldValue(IssueFieldId.LABELS_FIELD.id, labels)
+                .build();
+        try {
+            jiraRestClient.getIssueClient().updateIssue(issueKey, issueInput).get(timeout, TimeUnit.SECONDS);
+        } catch (Exception e) {
+            LOGGER.log(Level.WARNING, "jira rest client update labels error for issue "+issueKey, e);
```
lanwen: use formatter here too
`LOGGER.log(Level.WARNING, "jira rest client update labels error for issue {0}", issueKey, e);`

src/main/java/hudson/plugins/jira/JiraSession.java
```diff
@@ -66,6 +73,24 @@ public void addComment(String issueId, String comment,
         service.addComment(issueId, comment, groupVisibility, roleVisibility);
     }
 
+    public void addLabels(String issueId, List<String> labels) {
```
lanwen: please write some description to javadoc
some kind of
> Adds new labels to issue. Old labels remains untouched

src/main/java/hudson/plugins/jira/Updater.java
```diff
@@ -184,76 +227,89 @@ static void submitComments(
...
-                        comment.append("[").append(url.toExternalForm()).append("]");
+                comment.append(change.getMsg());
+                String revision = getRevision(change);
+                if (revision != null) {
+                    URL url = null;
+                    if (repoBrowser != null) {
+                        try {
+                            url = repoBrowser.getChangeSetLink(change);
+                        } catch (IOException e) {
+                            LOGGER.warning("Failed to calculate SCM repository browser link " + e.getMessage());
```
lanwen: think we should log stacktrace at least to the fine lvl

src/main/java/hudson/plugins/jira/Updater.java
jan-zajic: ok, i will remove this line

src/main/java/hudson/plugins/jira/JiraChangeLogAnnotator.java
jan-zajic: i will change it

src/main/java/hudson/plugins/jira/JiraIssueUpdater.java
jan-zajic: ok

src/test/java/hudson/plugins/jira/deprecated/DeprecatedJiraBuildAction.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    public String getUrlName() {
+        return "jira";
+    }
+
+    /**
+     * Finds {@link JiraIssue} whose ID matches the given one.
+     */
+    public JiraIssue getIssue(String id) {
```
lanwen: please add annotation `@CheckForNull`

warden:
Just my thoughts: maybe it would be nice to move the whole SCM logic (getChanges, etc...) and workarounds to a separate class to clean up JiraIssueUpdater and Updater.

src/main/java/hudson/plugins/jira/JiraRestService.java
jan-zajic: NO, this is NOT possible with wretched java logging api. You have options:
* Log a message, with one object parameter. 
* Log a message, with an array of object arguments. 
* Log a message, with associated Throwable information. (Note that the thrown argument is stored in the LogRecord thrown property, rather than the LogRecord parameters property. Thus is it processed specially by output Formatters and is not treated as a formatting parameter to the LogRecord message property. )

lanwen: is there slf4j logger in classpath?

src/main/java/hudson/plugins/jira/JiraSession.java
jan-zajic: Ok, i add something like this

/**
     * Adds new labels to the existing issue.
     * Old labels remains untouched.
     *
     * @param issueId Jira issue ID like "MNG-1235".
     * @param labels New labels to add.
     */

src/main/java/hudson/plugins/jira/Updater.java
jan-zajic: I think so.

src/test/java/hudson/plugins/jira/deprecated/DeprecatedJiraBuildAction.java
jan-zajic: No, this is only fixated class that reflect state before this PR changes, for testing purposes (see javadoc).

src/main/java/hudson/plugins/jira/JiraRestService.java
jan-zajic: Yes is, personally i always prefer slf4j, but i don't think it be happily to add two different loggers in one class or to change logging in whole class in this pull request (unrelated changes - maybe change logging in whole project later if someone have time to do).

warden: Agreed, this can be a part of next refactor.

src/test/java/hudson/plugins/jira/deprecated/DeprecatedJiraBuildAction.java
lanwen: oh, i see now, sorry

src/main/java/hudson/plugins/jira/JiraRestService.java
lanwen: ok, thanks for investigation

lanwen:
ok for me after some log changes

jan-zajic:
ok, i pushed fixes.

jan-zajic:
so, @warden, i think we are done? ;)

warden:
I think I'm good with merging that in.
One last thing: can you add a usage example of workflow configurations?
Please create an EXAMPLES.md file and add some snippets there, I think that will be useful for newcomers.

jan-zajic:
@warden I'm created file COMPATIBILITY.md with some explanation. Please check it after me and tell what would you like to change or add.

jan-zajic:
@warden do you mean like that?

oleg-nenashev:
@jan-zajic @warden 
FYI Workflow is being renamed to Pipeline. Maybe makes sense to cleanup UIs and document pages

PR #77, opened 2016-01-06
console message improvement for VersionReleaser

deki:
use resource bundle instead of string constant, also print a message on success)


src/test/java/hudson/plugins/jira/JiraReleaseVersionUpdaterTest.java
```diff
@@ -47,6 +48,7 @@ public void createMocks() {
 		env = mock(EnvVars.class);
 		project = mock(AbstractProject.class);
 		site = mock(JiraSite.class);
+      logger = mock(PrintStream.class);
```
warden: Please fix indentation in the test class

deki:
tabs and spaces are mixed in several classes, what is preferred?

warden:
Please use spaces as the preferred way, thanks!

deki:
Ok I already used spaces in my commit, so I will replace the tabs in the other lines.

PR #78, opened 2016-01-07
console message improvement for JiraVersionCreator

deki:
Similar to #77 - use resource bundle instead of string constant, also print a message on success


warden:
:+1:
In this case, because the two commits are targeting the same change , it is better to squash them.

PR #79, opened 2016-01-10
Add JiraEnvironmentVariableBuilder

sfhardman:
This is a new build step that adds JIRA info to the build environment variables.  

The main use case is to add this to the build, then use the resulting JIRA_ISSUES variable (extracted from the commit log) in the JQL selector for "Progress JIRA issues by workflow action" or "Move issues matching JQL to the specified version".  This supports an automated workflow where issues get picked up and modified automatically by being referenced in the commit log.

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/jira-plugin/79)

### DIFF
--- a/src/main/resources/hudson/plugins/jira/Messages.properties
+++ b/src/main/resources/hudson/plugins/jira/Messages.properties
@@ -26,3 +26,6 @@ VersionReleaser.AlreadyReleased=[JIRA] The version {0} is already released in pr
 VersionReleaser.MarkingReleased=[JIRA] Marking version {0} in project {1} as released.
 JiraVersionCreator.VersionExists=[JIRA] A version with name {0} already exists in project {1}, so nothing to do.
 JiraVersionCreator.CreatingVersion=[JIRA] Creating version {0} in project {1}.
+JiraEnvironmentVariableBuilder.DisplayName=Add JIRA related environment variables to build
+JiraEnvironmentVariableBuilder.NoJiraSite=[JIRA] No JIRA site is configured for this project. This must be a project configuration error
+JiraEnvironmentVariableBuilder.Updating=[JIRA] Setting {0} to {1}.
--- a/src/main/resources/hudson/plugins/jira/Messages.properties
+++ b/src/main/resources/hudson/plugins/jira/Messages.properties
@@ -26,6 +26,6 @@ VersionReleaser.AlreadyReleased=[JIRA] The version {0} is already released in pr
 VersionReleaser.MarkingReleased=[JIRA] Marking version {0} in project {1} as released.
 JiraVersionCreator.VersionExists=[JIRA] A version with name {0} already exists in project {1}, so nothing to do.
 JiraVersionCreator.CreatingVersion=[JIRA] Creating version {0} in project {1}.
-JiraEnvironmentVariableBuilder.DisplayName=Add JIRA related environment variables to build
+JiraEnvironmentVariableBuilder.DisplayName=JIRA: Add related environment variables to build
 JiraEnvironmentVariableBuilder.NoJiraSite=[JIRA] No JIRA site is configured for this project. This must be a project configuration error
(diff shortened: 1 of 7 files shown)


src/test/java/hudson/plugins/jira/JiraEnvironmentVariableBuilderTest.java
```diff
@@ -0,0 +1,114 @@
...
+    @Test
+    public void testIssueSelectorDefaultsToDefault() {
+        final JiraEnvironmentVariableBuilder builder = new JiraEnvironmentVariableBuilder(null);
+        assertThat(builder.getIssueSelector(), instanceOf(DefaultUpdaterIssueSelector.class));
+    }
+
+    @Test
+    public void testSetIssueSelectorPersists() {
+        final JiraEnvironmentVariableBuilder builder = new JiraEnvironmentVariableBuilder(issueSelector);
+        assertThat(builder.getIssueSelector(), is(issueSelector));
```
warden: Not sure what happens here.
testIssueSelectorDefaultsToDefault tests the default selector and assigns it to the test class variable. And then this test checks if that was assigned correctly? Also, if I remember correct, junit tests are not guaranteed to execute in the order they are named, so this may occasionally fail.
(but I might be wrong (R) ;-) )

sfhardman: Java newbie so may have got this wrong, but testIssueSelectorDefaultsToDefault is testing that when JiraEnvironmentVariableBuilder is created with a null selector it creates an instance of the default selector.  testSetIssueSelectorPersists is testing that if a selector is passed in it is used as the selector.  AFAIK the tests are independent.  

This bit is pretty much a clone of JiraIssueUpdaterTest.java but using mockito instead of a defined subclass of UpdaterIssueSelector

warden: :+1:

warden:
Could you ensure Jenkins Workflow compatibility?
This will mean probably implementing SimpleBuildStep in the Builder.
See [here](https://github.com/jenkinsci/workflow-plugin/blob/master/COMPATIBILITY.md) for some info.

sfhardman:
Any chance the workflow compatibility change for this builder could be parked pending resolution of https://issues.jenkins-ci.org/browse/JENKINS-29537?  The available [workarounds](http://stackoverflow.com/questions/31326286/can-a-workflow-step-access-environment-variables-provided-by-an-environmentcontr) don't seem very elegant

warden:
That makes sense, please create a ticket for that without assignee. 
The first workaround seems the cleanest at first sight. I'm pretty sure that a lot of people would see this functionality compatible with workflow, so whenever you feel like, PRs will be welcome :)

warden:
Also see #81.

src/main/resources/hudson/plugins/jira/Messages.properties
```diff
@@ -26,3 +26,6 @@ VersionReleaser.AlreadyReleased=[JIRA] The version {0} is already released in pr
 VersionReleaser.MarkingReleased=[JIRA] Marking version {0} in project {1} as released.
 JiraVersionCreator.VersionExists=[JIRA] A version with name {0} already exists in project {1}, so nothing to do.
 JiraVersionCreator.CreatingVersion=[JIRA] Creating version {0} in project {1}.
+JiraEnvironmentVariableBuilder.DisplayName=Add JIRA related environment variables to build
```
warden: Change to below, to fix dropdown ordering:
"JIRA: Add related environment variables to build"

src/test/java/hudson/plugins/jira/JiraEnvironmentVariableBuilderTest.java
warden: !completed

sfhardman:
Logged https://issues.jenkins-ci.org/browse/JENKINS-32504

PR #81, opened 2016-01-18
Custom Workflow Pipeline steps

jan-zajic:
I create a specialized workflow steps that make it easy to use actions from Jenkins workflow plugin.
* search for an issues based on jql query
* post a arbitrary comment to defined issue

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/jira-plugin/81)

### DIFF
--- a/COMPATIBILITY.md
+++ b/COMPATIBILITY.md
@@ -1,4 +1,5 @@
-# Plugin Compatibility with [Workflow](https://github.com/jenkinsci/workflow-plugin)
+# Plugin Compatibility with [Pipeline](https://github.com/jenkinsci/workflow-plugin)
+(formerly known as workflow plugin)
 
 This document captures the status of features to be compatible or incompatible.
 
@@ -28,16 +29,36 @@ You can add some labels to issue in jira:
             labels: [ "$version", "jenkins" ]])            
 ```
 
+##SearchIssuesStep
+
+Custom pipeline step (see [step-api](https://github.com/jenkinsci/workflow-plugin/blob/master/step-api/README.md)) that allow to search by jql query directly from workflow.
+
+usage:
+```groovy
+node {
+    List<String> issueKeys = jiraSearch(jql: "project = EX and labels = 'jenkins' and labels = '${version}'")	
+}
+```
+
+##CommentStep
+
+Interface for Pipeline job types that simply want to post a comment e.g.
+```groovy
+node {
+    jiraComment(issueKey: "EX-111", body: "Job '${env.JOB_NAME}' (${env.BUILD_NUMBER}) builded. Please go to ${env.BUILD_URL}.")
+}
+```
+
 ## Other features
 
-Some features are currently not supported in workflow.
-If you are adding new features please make sure that they support Jenkins Workflow Plugin.
+Some features are currently not supported in pipeline.
+If you are adding new features please make sure that they support Jenkins pipeline Plugin.
 See [here](https://github.com/jenkinsci/workflow-plugin/blob/master/COMPATIBILITY.md) for some information.
 See [here](https://github.com/jenkinsci/workflow-plugin/blob/master/basic-steps/CORE-STEPS.md) for more information how core jenkins steps integrate with workflow jobs.
 
 Running a notifiers is trickier since normally a flow in progress has no status yet, unlike a freestyle project whose status is determined before the notifier is called (never supported).
 So notifiers will never be implemented as you can use the catchError step and run jira action manually.
-I'm going to create a special workflow steps to replace this notifiers in future.
+I'm going to create a special pipeline steps to replace this notifiers in future.
 
 Other builders will be supported in future (not supported yet).
 
--- a/pom.xml
+++ b/pom.xml
@@ -65,6 +65,11 @@
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-surefire-plugin</artifactId>
           <version>2.12</version>
+          <configuration>
+            <forkCount>3</forkCount>
+            <reuseForks>true</reuseForks>
+            <argLine>-Xmx1024m</argLine>
+          </configuration>
         </plugin>
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
@@ -284,6 +289,12 @@
       <artifactId>jackson-databind</artifactId>
       <version>2.6.0-rc3</version>
     </dependency>
+	<dependency>
+	  <artifactId>workflow-step-api</artifactId>
+	  <groupId>org.jenkins-ci.plugins.workflow</groupId>
+	  <version>1.12</version>
+	  <optional>true</optional>
+	</dependency>
 
     <!-- test dependencies -->
     <dependency>
@@ -326,6 +337,13 @@
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <artifactId>workflow-job</artifactId>
       <version>1.12</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-step-api</artifactId>
+      <version>1.12</version>
+      <classifier>tests</classifier>
       <scope>test</scope>
     </dependency>
   </dependencies>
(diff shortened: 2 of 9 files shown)


jan-zajic:
It is polite to say that this change forces optional dependency on workflow-step-api.

jan-zajic:
@oleg-nenashev, @lanwen any comments on this? Thank you for your attention.

COMPATIBILITY.md
```diff
@@ -28,6 +28,26 @@ You can add some labels to issue in jira:
             labels: [ "$version", "jenkins" ]])            
 ```
 
+##SearchIssuesStep
+
+Workflow custom pipeline step (see [step-api](https://github.com/jenkinsci/workflow-plugin/blob/master/step-api/README.md)) that allow to search by jql query directly from workflow.
```
oleg-nenashev: Now it's Pipeline. See the thread here: https://issues.jenkins-ci.org/browse/JENKINS-31153

src/main/java/hudson/plugins/jira/workflow/CommentStep.java
```diff
@@ -0,0 +1,99 @@
...
+
+    public String getIssueKey() {
+        return issueKey;
+    }
+
+    public String getBody() {
+        return body;
+    }
+
+    @Extension
```
oleg-nenashev: Should be optional since it comes from the optional plugin
<code>@Extension(optional=true)</code>

src/main/java/hudson/plugins/jira/workflow/CommentStep.java
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        @Override
+        public String getFunctionName() {
+            return "jiraComment";
+        }
+
+        @Override
+        public String getDisplayName() {
+            return "Jira comment";
```
oleg-nenashev: Maybe "Add a comment to JIRA issue"?

src/main/java/hudson/plugins/jira/workflow/CommentStep.java
```diff
@@ -0,0 +1,99 @@
...
+import hudson.plugins.jira.JiraSession;
+import hudson.plugins.jira.JiraSite;
+import hudson.plugins.jira.Messages;
+
+/**
+ * Simple add comment step.
+ *
+ * @author jan zajic
+ */
+public class CommentStep extends AbstractStepImpl {
```
oleg-nenashev: Nice2have: <code>@since TODO</code>

src/main/java/hudson/plugins/jira/workflow/SearchIssuesStep.java
```diff
@@ -0,0 +1,99 @@
...
+    @DataBoundConstructor
+    public SearchIssuesStep(@Nonnull String jql) {
+        this.jql = jql;
+    }
+
+    public String getJql() {
+        return jql;
+    }
+
+    @Extension
```
oleg-nenashev: same as above

src/main/java/hudson/plugins/jira/workflow/CommentStep.java
warden: :+1:

oleg-nenashev: Or even "JIRA: Add comment to the issue". It would allow to workaround workflow sorting issues

src/main/resources/hudson/plugins/jira/workflow/SearchIssuesStep/config.jelly
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
+    <f:entry field="jql" title="Jql query">
```
warden: JQL

src/main/java/hudson/plugins/jira/workflow/CommentStep.java
warden: @oleg-nenashev that's a cool idea! I'll file a ticket for re-writing other UIs selects.

src/main/java/hudson/plugins/jira/workflow/CommentStep.java
```diff
@@ -0,0 +1,99 @@
...
+            return "Jira comment";
+        }
+    }
+
+    /**
+     * @author jan zajic
+     */
+    public static class CommentStepExecution extends AbstractSynchronousNonBlockingStepExecution<Void> {
+
+        private static final long serialVersionUID = 1L;
```
warden: I think it's better to not define this or define a generated value.

src/main/java/hudson/plugins/jira/workflow/CommentStep.java
lanwen: Propose to use `Messages.properties` to be localizable

src/main/java/hudson/plugins/jira/workflow/CommentStep.java
lanwen: first version should be 1L, then if class will be changed - should use another version

lanwen: first version should be 1L, then if class will be changed - should use another version

lanwen: first version should be 1L, then if class will be changed - should use another version

src/main/java/hudson/plugins/jira/workflow/CommentStep.java
```diff
@@ -0,0 +1,99 @@
+package hudson.plugins.jira.workflow;
```
lanwen: so the package name should be `pipeline` too

src/main/java/hudson/plugins/jira/workflow/CommentStep.java
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * @author jan zajic
+     */
+    public static class CommentStepExecution extends AbstractSynchronousNonBlockingStepExecution<Void> {
+
+        private static final long serialVersionUID = 1L;
+
+        @Inject
+        private transient CommentStep step;
```
lanwen: wow, how does this works without `@Extension` on class?

https://groups.google.com/forum/#!topic/jenkinsci-dev/ZutbuAfl10s

src/test/java/hudson/plugins/jira/workflow/SearchIssuesStepTest.java
```diff
@@ -0,0 +1,105 @@
...
+import hudson.plugins.jira.JiraSite;
+import hudson.plugins.jira.workflow.SearchIssuesStep.SearchStepExecution;
+
+public class SearchIssuesStepTest {
+
+    @Rule
+    public JenkinsRule j = new JenkinsRule();
+
+    @Inject
+    SearchIssuesStep.DescriptorImpl d;
```
warden: d -> descriptor

src/main/java/hudson/plugins/jira/workflow/CommentStep.java
jan-zajic: I followed these guidelines https://github.com/jenkinsci/workflow-plugin/blob/master/step-api/README.md.

src/main/java/hudson/plugins/jira/workflow/CommentStep.java
jan-zajic: see org.jenkinsci.plugins.workflow.steps.AbstractStepImpl.start

jan-zajic:
@oleg-nenashev, @lanwen, @warden i fix your comments. It looks like impossible to run jenkins tests with JenkinsRule on travis-ci, probably some resources problem? Jenkins build it fine..

warden:
See the error message:
```
Running hudson.plugins.jira.pipeline.SearchIssuesStepTest
Killed
```
This might be to a number of reasons. Try experimenting with the hints [from SO](http://stackoverflow.com/questions/23260057/the-forked-vm-terminated-without-saying-properly-goodbye-vm-crash-or-system-exi). I am guessing this might be because of some memory limit, so try setting surefire-plugin configuration.

warden:
See #85 for fix

PR #84, opened 2016-01-18
JENKINS-32106: fixed UNKNOWN in release notes

deki:
[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/jira-plugin/84)

### DIFF
--- a/src/main/java/hudson/plugins/jira/JiraSite.java
+++ b/src/main/java/hudson/plugins/jira/JiraSite.java
@@ -473,13 +473,6 @@ public class JiraSite extends AbstractDescribableImpl<JiraSite> {
         }
 
         List<Issue> issues = session.getIssuesWithFixVersion(projectKey, versionName, filter);
-        List<IssueType> types = session.getIssueTypes();
-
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/plugins/jira/JiraSite.java
```diff
@@ -491,13 +484,7 @@ public String getReleaseNotesForFixVersion(String projectKey, String versionName
             String key = issue.getKey();
             String summary = issue.getSummary();
             String status = issue.getStatus().getName();
-            String type = "UNKNOWN";
```
warden: This doesn't look like it fixes the issue. Moreover it looks like it will not fallback to UNKNOWN in case the issueType is not known...
Also, this needs a test.

deki:
Currently the issue type is always UNKNOWN as the typeNameMap has a numeric key but is queried with the String issueTypeName. Seems that this was introduced when switching from SOAP to REST API. From my point of view the issueTypeName is always set and can be used directly (same is done for status). I can create a test and reintroduce the default UNKNOWN if you feel better with it...

PR #91, opened 2016-03-07
Bugfix: Do not fail if labels list is null

ercpe:
For some reason, the labels list can be null. Thus, the JIRA plugin throwed a NPE and stored the changes for the next run *after* the comment has been added to the first issue. This caused the same comment added to the issues over and over at each run.

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/jira-plugin/91)


warden:
Fixed in #92

PR #94, opened 2016-04-04
[JENKINS-33996] added new Pipeline step that find all JIRA issues ref…

escoem:
added new Pipeline step that find all JIRA issues referenced by the commit logs of current build

Please @jglick @olamy @warden would you mind to review this PR?

Also @reviewbybees

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/jira-plugin/94)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -21,6 +21,7 @@
     <mockito.version>1.10.19</mockito.version>
     <powermock.version>1.6.3</powermock.version>
     <jira-rest-client.version>3.0.0</jira-rest-client.version>
+    <pipeline.version>1.12</pipeline.version>
 
     <!-- security -->
     <owasp.version>1.3.1</owasp.version>
@@ -291,13 +292,13 @@
 	<dependency>
 	  <artifactId>workflow-step-api</artifactId>
 	  <groupId>org.jenkins-ci.plugins.workflow</groupId>
-	  <version>1.12</version>
+	  <version>${pipeline.version}</version>
 	  <optional>true</optional>
 	</dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <artifactId>workflow-job</artifactId>
-      <version>1.12</version>
+      <version>${pipeline.version}</version>
       <optional>true</optional>
     </dependency>
     
@@ -335,13 +336,13 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <artifactId>workflow-aggregator</artifactId>
-      <version>1.12</version>
+      <version>${pipeline.version}</version>
       <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <artifactId>workflow-step-api</artifactId>
-      <version>1.12</version>
+      <version>${pipeline.version}</version>
       <classifier>tests</classifier>
       <scope>test</scope>
     </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -145,7 +145,6 @@
         </executions>
       </plugin>
 
-
       <plugin>
         <groupId>org.eluder.coveralls</groupId>
         <artifactId>coveralls-maven-plugin</artifactId>
@@ -295,7 +294,13 @@
 	  <version>1.12</version>
 	  <optional>true</optional>
 	</dependency>
-
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-job</artifactId>
+      <version>1.12</version>
+      <optional>true</optional>
+    </dependency>
+    
     <!-- test dependencies -->
     <dependency>
       <groupId>org.mockito</groupId>
@@ -329,13 +334,7 @@
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
-      <artifactId>workflow-scm-step</artifactId>
-      <version>1.12</version>
-      <scope>test</scope>
-    </dependency>
-   	<dependency>
-      <groupId>org.jenkins-ci.plugins.workflow</groupId>
-      <artifactId>workflow-job</artifactId>
+      <artifactId>workflow-aggregator</artifactId>
       <version>1.12</version>
       <scope>test</scope>
     </dependency>
--- a/src/main/java/hudson/plugins/jira/pipeline/FindIssuesAtCommitLogsStep.java
+++ b/src/main/java/hudson/plugins/jira/pipeline/FindIssuesAtCommitLogsStep.java
@@ -22,6 +22,7 @@ import com.google.inject.Inject;
 
 import hudson.Extension;
 import hudson.model.TaskListener;
+import hudson.plugins.jira.JiraSite;
 import hudson.plugins.jira.Messages;
 import hudson.scm.ChangeLogSet;
 import hudson.scm.ChangeLogSet.Entry;
@@ -33,8 +34,7 @@ import hudson.scm.ChangeLogSet.Entry;
  */
 public class FindIssuesAtCommitLogsStep extends AbstractStepImpl {
 
-    public static final String DEFAULT_ISSUE_PATTERN = "((?<!([A-Z]{1,10})-?)[A-Z]+-\\d+)";
-    protected String issuePattern = DEFAULT_ISSUE_PATTERN;
+    protected String issuePattern;
     
     @DataBoundConstructor
     public FindIssuesAtCommitLogsStep() {
@@ -42,11 +42,11 @@ public class FindIssuesAtCommitLogsStep extends AbstractStepImpl {
     
     @DataBoundSetter
     public void setIssuePattern(@Nonnull String issuePattern) {
-        this.issuePattern = StringUtils.isBlank(issuePattern) ? DEFAULT_ISSUE_PATTERN : issuePattern;
+        this.issuePattern = issuePattern;
     }
     
     public String getIssuePattern() {
-        return issuePattern;
+        return StringUtils.isBlank(issuePattern) ? JiraSite.DEFAULT_ISSUE_PATTERN.toString() : issuePattern;
     }
     
     @Extension(optional = true)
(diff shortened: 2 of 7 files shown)


src/main/java/hudson/plugins/jira/pipeline/FindIssuesAtCommitLogsStep.java
```diff
@@ -0,0 +1,109 @@
...
+import hudson.scm.ChangeLogSet.Entry;
+
+/**
+ * Find all issues referenced by the commit logs of current build
+ *
+ * @author emilio escobar
+ */
+public class FindIssuesAtCommitLogsStep extends AbstractStepImpl {
+
+    public static final String DEFAULT_ISSUE_PATTERN = "((?<!([A-Z]{1,10})-?)[A-Z]+-\\d+)";
```
warden: This is in JiraSite.DEFAULT_ISSUE_PATTERN

pom.xml
```diff
@@ -295,7 +294,13 @@
 	  <version>1.12</version>
 	  <optional>true</optional>
 	</dependency>
-
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-job</artifactId>
+      <version>1.12</version>
```
warden: Better to move this along with workflow-aggregator to workflow.version property

src/main/resources/hudson/plugins/jira/pipeline/FindIssuesAtCommitLogsStep/config.jelly
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
warden: This is not needed as IssuePattern is defined on JiraSite level

warden:
So... how is this different from IssueSelectorStep ?

src/main/java/hudson/plugins/jira/pipeline/FindIssuesAtCommitLogsStep.java
```diff
@@ -0,0 +1,109 @@
...
+    @Extension(optional = true)
+    public static final class DescriptorImpl extends AbstractStepDescriptorImpl {
+        
+        public DescriptorImpl() {
+            super(FindIssuesAtCommitLogsStepExecution.class);
+        }
+
+        @Override
+        public String getFunctionName() {
+            return "jiraFindIssuesAtCommitLogs";
```
amuniz: what about `jiraFindInChangeSets`? Juts a bit shorter.

src/main/java/hudson/plugins/jira/pipeline/FindIssuesAtCommitLogsStep.java
```diff
@@ -0,0 +1,109 @@
...
+    
+    public static class FindIssuesAtCommitLogsStepExecution extends AbstractSynchronousNonBlockingStepExecution<List<String>> {
+
+        private static final long serialVersionUID = -6080152002390517795L;
+
+        @Inject
+        private transient FindIssuesAtCommitLogsStep step;
+
+        @StepContextParameter
+        private transient TaskListener listener;
```
amuniz: This is unused.

src/main/resources/hudson/plugins/jira/Messages.properties
```diff
@@ -33,3 +33,4 @@ CommentStep.Descriptor.DisplayName=JIRA: Add a comment to issue(s)
 SearchIssuesStep.Descriptor.DisplayName=JIRA: Search issues
 IssueSelectorStep.Descriptor.DisplayName=JIRA: Issue selector
 JiraCreateIssueNotifier.DisplayName=JIRA: Create issue
+FindIssuesAtCommitLogsStep.Descriptor.DisplayName=JIRA: find all issues referenced by the commit logs of this build
```
amuniz: I would remove `JIRA:` as it is going to be shown as `jiraFindIssuesAtCommitLogs: JIRA: find all issues referenced by the commit logs of this build`. The repeated `:` looks weird (although all other steps in this plugin seems to be using the same display message pattern...)

src/test/java/hudson/plugins/jira/pipeline/FindIssuesAtCommitLogsStepTest.java
```diff
@@ -0,0 +1,32 @@
...
+import org.junit.Before;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.jvnet.hudson.test.JenkinsRule;
+
+import com.google.inject.Inject;
+
+public class FindIssuesAtCommitLogsStepTest {
+
+    @ClassRule
```
amuniz: Use `@Rule` unless you really need this to be a class rule.

src/test/java/hudson/plugins/jira/pipeline/FindIssuesAtCommitLogsStepTest.java
```diff
@@ -0,0 +1,32 @@
...
+    public static JenkinsRule jenkinsRule = new JenkinsRule();
+    
+    @Inject
+    FindIssuesAtCommitLogsStep.DescriptorImpl descriptor;
```
amuniz: This is not used.

src/test/java/hudson/plugins/jira/pipeline/FindIssuesAtCommitLogsStepTest.java
```diff
@@ -0,0 +1,32 @@
...
+
+    @Before
+    public void setUp() {
+        jenkinsRule.getInstance().getInjector().injectMembers(this);
```
amuniz: Not needed AFAICT

amuniz:
Looks good. But I think a simple test would be in place, perhaps if you extract the logic behind `Run` (which is basically the `for` loop inside `FindIssuesAtCommitLogsStepExecution.run`) in a utility method so you can call it without the need to start Jenkins makes the tests easier to write.

src/main/resources/hudson/plugins/jira/Messages.properties
warden: This pattern is to group all JIRA related options in the dropdown select menu..

src/main/java/hudson/plugins/jira/pipeline/FindIssuesAtCommitLogsStep.java
escoem: Ok, I will use JiraSite.DEFAULT_ISSUE_PATTERN and remove the other one

pom.xml
escoem: Ok, pipeline.version at properties

src/main/resources/hudson/plugins/jira/pipeline/FindIssuesAtCommitLogsStep/config.jelly
escoem: Yes because I want to offer the possibility to overwrite it at run time

escoem:
Hi @warden @amuniz I have replied almost all of your comments and also committed the changes.

src/main/resources/hudson/plugins/jira/Messages.properties
amuniz: No, they are shown all consecutive because all step names start with `jira`, not because of the `JIRA: ` fragment in the description. Anyway, I'm nitpicking, you can omit my comment.

amuniz: I know what you meant now. In previous Workflow versions only the description was shown in the dropdown, true. But now (due to [this](https://github.com/jenkinsci/workflow-plugin/commit/577f2acb256c0c91eaa81abe3177b4248bbaa059)) the elements in that dropdown list are ordered by the step function name. This was introduced in `workflow:1.11`, so as long as you depending on 1.12 the `JIRA:` fragment could be removed safely.

src/main/java/hudson/plugins/jira/pipeline/FindIssuesAtCommitLogsStep.java
```diff
@@ -0,0 +1,107 @@
...
+
+        @Override
+        public String getDisplayName() {
+            return Messages.FindIssuesAtCommitLogsStep_Descriptor_DisplayName();
+        }
+    }
+    
+    public static class FindIssuesAtCommitLogsStepExecution extends AbstractSynchronousNonBlockingStepExecution<List<String>> {
+
+        private static final long serialVersionUID = -6080152002390517795L;
```
amuniz: This should be `1L` (as the first version of this class).

src/main/resources/hudson/plugins/jira/pipeline/FindIssuesAtCommitLogsStep/config.jelly
warden: What would you want to overwrite issuePattern in runtime? Can you describe the usecase?

src/main/resources/hudson/plugins/jira/Messages.properties
warden: Ah, that's interesting, but I'd like to group all dropdown options together (i.e. not only workflow ones). Can this be done?

amuniz: >  not only workflow ones

All steps in that list are Pipeline steps, what do you mean?

src/main/resources/hudson/plugins/jira/pipeline/FindIssuesAtCommitLogsStep/config.jelly
amuniz: I guess you mean [this](https://github.com/escoem/jira-plugin/blob/JENKINS-33996/src/main/resources/hudson/plugins/jira/JiraSite/config.jelly#L17), right?

This new jelly is defining a simplified configuration (actually just offering one single configuration option), so JiraSite jelly would not re-usable here.

This simplified jelly is only used by Pipeline Snippet Generator.

escoem:
Hi @warden do you need anything else? Any other change or clarification?

escoem:
Closed in favour of https://github.com/jenkinsci/jira-plugin/pull/96

PR #95, opened 2016-04-27
JENKINS-24207: Add new selector for Perforce/P4 that allows to update JIRA issues based on jobs

J-cztery:
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/jira-plugin/95)


warden:
Nice! Can I ask you to create JIRA ticket for this in issues.jenkins-ci.org and reference that JIRA number in the commit message? That helps in the release notes :)

src/main/java/hudson/plugins/jira/selector/p4/JobIssueSelector.java
```diff
@@ -0,0 +1,18 @@
...
+import hudson.model.Run;
+import hudson.model.TaskListener;
+import hudson.plugins.jira.JiraSite;
+import hudson.plugins.jira.selector.AbstractIssueSelector;
+
+public class JobIssueSelector extends AbstractIssueSelector {
+
+    @Override
+    public Set<String> findIssueIds(Run<?, ?> run, JiraSite site, TaskListener listener) {
+        // TODO Auto-generated method stub
```
warden: If this is not implemented yet, do we need to have it?

J-cztery: No, definitely not. I just wanted to have your opinion on the direction. It got late last night so i had to stop. I will either implement p4 plugin or remove dependencies on it and the selector.

warden: Great, sounds good, thanks!

src/main/java/hudson/plugins/jira/selector/perforce/JobIssueSelector.java
```diff
@@ -0,0 +1,152 @@
...
+        return ids;
+    }
+
+    private void printInfoAboutFoundIssues(Set<String> ids, TaskListener listener) {
+       if (ids.isEmpty()){
+           listener.getLogger().println("JIRA: Found no JIRA issues.");
+       }else{
+           listener.getLogger().print("JIRA: The following JIRA issues will be updated: ");
+           Iterator<String> it = ids.iterator();
+           while(it.hasNext()){
```
warden: You could use StringUtils.join()

J-cztery:
There already was a JIRA issue. Let me know if you have any other comments. I will squash the commits once it is final.

warden:
There's one thing I don't like - the static methods in DefaultIssueSelector and using them in the JobIssueSelector in that way.. Maybe you could move them AbstractIssueSelector and make non-static?

src/main/java/hudson/plugins/jira/selector/jobs/JobIssueSelector.java
```diff
@@ -0,0 +1,62 @@
...
+
+import hudson.model.AbstractBuild;
+import hudson.model.Run;
+import hudson.model.TaskListener;
+import hudson.model.AbstractBuild.DependencyChange;
+import hudson.plugins.jira.JiraSite;
+import hudson.plugins.jira.RunScmChangeExtractor;
+import hudson.plugins.jira.selector.AbstractIssueSelector;
+import hudson.plugins.jira.selector.DefaultIssueSelector;
+
```
warden: Please also add JavaDoc describing this class

J-cztery:
@warden AbstractIssueSelector is not a good place for this methods. But I see what you are coming from. Please have a look at the changes i made (I think they are what you wanted but they are in DefaultIssueSelector rather than in AbstractIssueSelector) and see if you like them or not. 
The idea is that DefaultIssueSelector contains basic logic, i.e. take issues from parameters, carried over, dependent builds and from current builds. Any piece of this behaviour can be overridden. 

Added javadoc. 

Added config.jelly resources for selectors, including DefaultIssueSelector.

Strings are now taken from properties..

src/main/java/hudson/plugins/jira/selector/DefaultIssueSelector.java
```diff
@@ -27,64 +29,46 @@
...
-        }
-        return ids;
-    }
-
-    /**
-     * @param pattern pattern to use to match issue ids
-     *            
-     */
-    protected static void findIssues(Run<?, ?> build, Set<String> ids, Pattern pattern, TaskListener listener) {
+    protected static void findIssues(Run<?, ?> build, Set<String> issueIds, Pattern pattern, TaskListener listener) {
```
warden: this static should not be needed here anymore :)

src/main/java/hudson/plugins/jira/selector/jobs/JobIssueSelector.java
```diff
@@ -0,0 +1,34 @@
+package hudson.plugins.jira.selector.jobs;
```
warden: Great changes, I like this now!
One thing only - as the "Job"s term are coming from Perforce and in Jenkins there is a similar term for a Job, I find this package name confusing. I propose to rename: hudson.plugins.jira.selector.jobs -> udson.plugins.jira.selector.perforce and move all classes there.
If you agree and this is done, I think we'll be good to merge!

src/main/java/hudson/plugins/jira/selector/jobs/perforce/PerforceJobIssueSelector.java
```diff
@@ -0,0 +1,79 @@
...
+ * Job selector for old Perforce SCM plugin
+ * 
+ * @author Jacek Tomaka
+ * @since 2.3
+ *
+ */
+public class PerforceJobIssueSelector extends JobIssueSelector {
+    private static final Logger LOGGER = Logger.getLogger(PerforceJobIssueSelector.class.getName());
+
+    @Extension(optional = true)
```
warden: not sure, but I think that @Extension on class level should be enough here

src/main/java/hudson/plugins/jira/selector/DefaultIssueSelector.java
J-cztery: If i remove it, i will break binary compatibility... 
Also, tests are using it... Tests can be reworked to not use this. But binary compatibility might be a bigger issue.

src/main/java/hudson/plugins/jira/selector/jobs/perforce/PerforceJobIssueSelector.java
J-cztery: Well, i copied what DefaultIssueSelector did.

src/main/java/hudson/plugins/jira/selector/DefaultIssueSelector.java
warden: True, that's a valid point, lets keep it as is then for now.

src/main/java/hudson/plugins/jira/selector/jobs/JobIssueSelector.java
J-cztery: Ok. Makes sense. I though there will be other job based SCMs so it would make sense but i can't find any. So let's apply YAGNI principle here.

src/main/java/hudson/plugins/jira/selector/perforce/PerforceJobIssueSelector.java
```diff
@@ -0,0 +1,78 @@
...
+import hudson.model.TaskListener;
+import hudson.plugins.jira.JiraSite;
+import hudson.plugins.jira.Messages;
+import hudson.plugins.jira.RunScmChangeExtractor;
+import hudson.plugins.jira.selector.AbstractIssueSelector;
+import hudson.plugins.perforce.PerforceChangeLogEntry;
+import hudson.scm.ChangeLogSet;
+import hudson.scm.ChangeLogSet.Entry;
+
+@Extension(optional = true)
```
warden: Double annotation, one is already at DescriptorImpl.

src/main/java/hudson/plugins/jira/selector/perforce/P4JobIssueSelector.java
```diff
@@ -0,0 +1,74 @@
...
+import hudson.model.Run;
+import hudson.model.TaskListener;
+import hudson.plugins.jira.JiraSite;
+import hudson.plugins.jira.Messages;
+import hudson.plugins.jira.RunScmChangeExtractor;
+import hudson.plugins.jira.selector.AbstractIssueSelector;
+import hudson.scm.ChangeLogSet;
+import hudson.scm.ChangeLogSet.Entry;
+
+@Extension(optional = true)
```
warden: same here

warden:
So I guess after removing the doubled annotations I'm okay with this one 👍

src/main/java/hudson/plugins/jira/selector/perforce/PerforceJobIssueSelector.java
J-cztery: I am not sure what you are referring to. These are two different classes. Each of them is marked as an extension. Feel free to modify that code if you do not like it. 
I do not have any more time to test if @Extension annotates nested classes automatically and documenation does not state it clearly.

J-cztery:
@warden this is how much i can afford to contribute my work to this change. At this time it is either take it or leave it. I am sure there are others willing to contribute the changes you require. 
Thank you for your useful comments and proposals!

src/main/java/hudson/plugins/jira/selector/perforce/PerforceJobIssueSelector.java
warden: Per documentation:
```
 * Marks a field, a method, or a class for automatic discovery, so that Hudson can locate
 * implementations of {@link ExtensionPoint}s automatically.
 *
 * <p>
 * (In contrast, in earlier Hudson, the registration was manual.)
 *
 * <p>
 * In a simplest case, put this on your class, and Hudson will create an instance of it
 * and register it to the appropriate {@link ExtensionList}.
 *
 * <p>
 * If you'd like Hudson to call
 * a factory method instead of a constructor, put this annotation on your static factory
 * method. Hudson will invoke it and if the method returns a non-null instance, it'll be
 * registered. The return type of the method is used to determine which {@link ExtensionList}
 * will get the instance.
 *
 * Finally, you can put this annotation on a static field if the field contains a reference
 * to an instance that you'd like to register.
 *
```

See DefaultIssueSelector where @Extension is only at DescriptorImpl level.

PR #96, opened 2016-04-29
Issue selector step improvement

escoem:
Improvement for IssueSelectorStep when you use pipeline snippet generator. Now you can select between all AbstractIssueSelector in order to get a better result of groovy snippet generation.

Also added a new issue selector that allows to override the default pattern.

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/jira-plugin/96)


escoem:
Hi @warden, I have closed the other PR (https://github.com/jenkinsci/jira-plugin/pull/94) and add this feature as an issue selector instead of a new step. I have also added the possibility of chose which issue selector implementation have to be used by issueSelectorStep at pipeline snippet generator.

@amuniz as you commented in the other PR, a new one would be great.

warden:
Thanks, it looks good to me, the only thing I can ask to improve the coverage so it at least doesn't decrease :)
@amuniz any comments from you?

src/main/java/hudson/plugins/jira/selector/ExplicitIssueSelector.java
```diff
@@ -13,24 +17,48 @@
...
 
     @CheckForNull
     private List<String> jiraIssueKeys;
+    private String issueKeys;
 
     @DataBoundConstructor
-    public ExplicitIssueSelector(String jiraIssueKeys) {
-        this(Lists.newArrayList(jiraIssueKeys.split(",")));
+    public ExplicitIssueSelector(String issueKeys) {
+        this.jiraIssueKeys = StringUtils.isNotBlank(issueKeys) ? Lists.newArrayList(issueKeys.split(",")) : Lists.<String>newArrayList();
```
amuniz: `Collections.emptyList()` would be more efficient.

src/main/java/hudson/plugins/jira/selector/JqlIssueSelector.java
```diff
@@ -27,6 +31,17 @@ public JqlIssueSelector(String jql) {
         this.jql = jql;
     }
 
+    public JqlIssueSelector(){
```
amuniz: Is this constructor needed?

amuniz:
Looks good to me 👍 (just some minor comments, and agree with @warden, some tests would be nice).

warden:
@escoem just heads up, that i'm planning to do release at the end of this week (probably on 8th May) if you want this to be part of it

escoem:
@warden merged with some comments. I have had to change JiraSite by Pattern in many methods because one static method was removed from DefaultIssueSelector in favour of addIssuesRecursive.

warden:
@escoem nice job!
Can you just squash your commits and reference JENKINS-33996 in your commit message?

src/main/java/hudson/plugins/jira/selector/ExplicitIssueSelector.java
warden: I agree :)

escoem:
hi @warden , I have tried but I don't know how to do it because the "merge". I have also tried a reword but didn't work. Maybe you can do from github UI? If not I will close this one and do a new PR :(

warden:
You should be able to do sth along the lines:
```
git fetch upstream
git rebase -i upstream/master
mark squashed + edit commit messages
git push -f
```

escoem:
hi @warden , that's what I did (I used git rebase -i HEAD~14 and also git rebase -i HEAD~14 --preserve-merge) and I have done (git rebase -i upstream/master) again but in both cases fails after the "mark squashed + edit commit messages" because the merge commit.

src/main/java/hudson/plugins/jira/selector/PatternIssueSelector.java
```diff
@@ -0,0 +1,56 @@
...
+    public PatternIssueSelector(){
+    }
+
+    @DataBoundSetter
+    public void setIssuePattern(String issuePattern) {
+        this.issuePattern = issuePattern;
+    }
+
+    public String getIssuePattern() {
+        return StringUtils.isBlank(issuePattern) ? JiraSite.DEFAULT_ISSUE_PATTERN.toString() : issuePattern;
```
warden: What is the usecase for using this PatternIssueSelector class?
I think normally you would use site.getIssuePattern() to get the pattern that is actually used by JIRA that you are linking against.

escoem: Because I haven't the JiraSite instance but is done there:
`
private Pattern getPattern(final JiraSite site) {
      return StringUtils.isBlank(issuePattern) ? site.getIssuePattern() : Pattern.compile(issuePattern);
}
`

warden: What I meant is what is PatternIssueSelector needed for as this issue pattern comes directly from DefaultIssueSelector.getIssuePattern ?

warden:
Not sure why you are rebasing against specific commit (HEAD~14) and not upstream/master ?
I don't know what was the merge commit (i'm assuming that was a merge from master, as I can see 11 commits when i do first command from below),
Have you tried removing last commit and then rebasing?
```
# revert last commit
git reset --hard HEAD~1
# rebase as usual
git rebase -i master
```

src/main/java/hudson/plugins/jira/selector/PatternIssueSelector.java
warden: For getting the JiraSite instance, maybe you could see how it's done in JiraChangeLogAnnotator.getSiteForProject().

escoem: Hi, for example if you need to use a more restrictive pattern in a specific pipeline but if you think that is something that shouldn't happen I can remove it (closing this pr and opening a new one only with the changes for snippet generator).

escoem:
Hi @warden I will wait for your reply to my [last comment](https://github.com/jenkinsci/jira-plugin/pull/96#discussion_r62995405) question

src/main/java/hudson/plugins/jira/selector/PatternIssueSelector.java
warden: Yes, I would propose to do so.
The reasoning is simple - JiraSite defines the pattern that is used against this specific JIRA instance, so redefining it here doesn't make too much sense.

escoem:
Closed in favour of https://github.com/jenkinsci/jira-plugin/pull/99

PR #97, opened 2016-05-01
Fix getting dependent issues

J-cztery:
As stated on the plugin Wiki: "By taking advantages of Jenkins' fingerprint feature, when your other projects that depend on this project pick up a build with a fix, those build numbers can also be recorded in JIRA". 
Selecting dependent issues currently does not work. This pull request is to fix it. 
This change depends on https://github.com/jenkinsci/jira-plugin/pull/95

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/jira-plugin/97)


J-cztery:
I am planning on adding a test case.

src/main/java/hudson/plugins/jira/selector/DefaultIssueSelector.java
```diff
@@ -128,7 +128,7 @@ protected void addIssuesFromCurrentBuild(Run<?, ?> build, JiraSite site, TaskLis
      */
     protected void addIssuesFromDependentBuilds(Run<?, ?> build, JiraSite site, TaskListener listener,
             Set<String> issueIds) {
-        for (DependencyChange depc : RunScmChangeExtractor.getDependencyChanges(build.getPreviousBuild()).values()) {
+        for (DependencyChange depc : RunScmChangeExtractor.getDependencyChanges(build).values()) {
```
warden: This is moving getPreviousBuild() to RunScmChangeExtractor

J-cztery: Previously it was getting dependency changes between previous build and ... previous build. 
Now it is taking dependency changes between current and previous build.

warden: 👍

PR #98, opened 2016-05-02
Feature: Possibility to add fix version to list instead of replacing whole list or specific version

mdenicki:
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/jira-plugin/98)

### DIFF
--- a/src/main/java/hudson/plugins/jira/JiraSession.java
+++ b/src/main/java/hudson/plugins/jira/JiraSession.java
@@ -251,6 +251,35 @@ public class JiraSession {
     }
 
     /**
+     * Adds the specified version to the fix version list of all issues matching the JQL.
+     *
+     * @param projectKey The JIRA Project
+     * @param version    The version to add
+     * @param query      The JQL Query
+     */
+    public void addFixVersion(String projectKey, String version, String query) {
+
+        Version newVersion = getVersionByName(projectKey, version);
+        if (newVersion == null) {
+            return;
+        }
+
+        LOGGER.fine("Fetching issues with JQL:" + query);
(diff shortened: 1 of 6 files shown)


src/main/java/hudson/plugins/jira/JiraSession.java
```diff
@@ -251,6 +251,35 @@ public void replaceFixVersion(String projectKey, String fromVersion, String toVe
...
+     * Adds the specified version to the fix version list of all issues matching the JQL.
+     *
+     * @param projectKey The JIRA Project
+     * @param version    The version to add
+     * @param query      The JQL Query
+     */
+    public void addFixVersion(String projectKey, String version, String query) {
+
+        Version newVersion = getVersionByName(projectKey, version);
+        if (newVersion == null) {
```
warden: Please add LOGGER.warn that the specified version was not found

src/main/java/hudson/plugins/jira/JiraSite.java
```diff
@@ -631,6 +631,23 @@ public void migrateIssuesToFixVersion(String projectKey, String versionName, Str
...
+     *
+     * @param projectKey
+     * @param versionName
+     * @param query
+     * @throws IOException
+     */
+    public void addFixVersionToIssue(String projectKey, String versionName, String query) throws IOException {
+        JiraSession session = getSession();
+        if (session == null) {
+            return;
```
warden: LOGGER.warn...

warden:
Thanks for nice new feature PR!
If you could just add those two loggers, I'll be happy to merge!

PR #100, opened 2016-05-13
JENKINS-33166 check for nonnull url inside constructor

escoem:
If you try to add and empty jira plugin configuration at Manage Jenkins, will fail with NPE.

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/jira-plugin/100)


warden:
Actually we shouldn't allow adding a JIRA site with an empty URL, what's the point ?:)
I think it is better to make the URL field required on the UI if possible.

PR #101, opened 2016-06-14
Add Jira Issue update step to pipeline

aatarasoff:
Similar as JiraIssueUpdateBuilder, but for pipeline

### DIFF
--- a/src/main/java/hudson/plugins/jira/pipeline/IssueUpdateStep.java
+++ b/src/main/java/hudson/plugins/jira/pipeline/IssueUpdateStep.java
@@ -1,0 +1,110 @@
+package hudson.plugins.jira.pipeline;
+
+import com.google.inject.Inject;
+import hudson.Extension;
+import hudson.Util;
+import hudson.model.Result;
+import hudson.model.Run;
+import hudson.model.TaskListener;
+import hudson.plugins.jira.JiraSite;
+import hudson.plugins.jira.Messages;
+import org.apache.commons.lang.StringUtils;
+import org.jenkinsci.plugins.workflow.steps.AbstractStepDescriptorImpl;
+import org.jenkinsci.plugins.workflow.steps.AbstractStepImpl;
+import org.jenkinsci.plugins.workflow.steps.AbstractSynchronousNonBlockingStepExecution;
+import org.jenkinsci.plugins.workflow.steps.StepContextParameter;
+import org.kohsuke.stapler.DataBoundConstructor;
+
+import javax.annotation.Nonnull;
+import java.io.IOException;
+
+/**
+ * Step for updating jira issue with workflow migration
+ *
+ * @author aatarasoff
+ */
+public class IssueUpdateStep extends AbstractStepImpl {
+    private final String jqlSearch;
+    private final String workflowActionName;
+    private final String comment;
+
+    @DataBoundConstructor
+    public IssueUpdateStep(@Nonnull String jqlSearch, @Nonnull String workflowActionName, String comment) {
+        this.jqlSearch = Util.fixEmptyAndTrim(jqlSearch);
+        this.workflowActionName = Util.fixEmptyAndTrim(workflowActionName);
+        this.comment = Util.fixEmptyAndTrim(comment);
+    }
+
+    public String getJqlSearch() {
+        return jqlSearch;
+    }
+
+    public String getWorkflowActionName() {
+        return workflowActionName;
+    }
+
+    public String getComment() {
+        return comment;
+    }
+
+    @Extension(optional = true)
+    public static final class DescriptorImpl extends AbstractStepDescriptorImpl {
+
+        public DescriptorImpl() {
+            super(IssueUpdateStep.StepExecution.class);
+        }
+
+        @Override
+        public String getFunctionName() {
+            return "jiraIssueUpdate";
+        }
+
+        @Override
+        public String getDisplayName() {
+            return Messages.IssueUpdateStep_Descriptor_DisplayName();
+        }
+    }
+
+    public static class StepExecution extends AbstractSynchronousNonBlockingStepExecution<Void> {
+
+        private static final long serialVersionUID = 1L;
+
+        @Inject
+        private transient IssueUpdateStep step;
+
+        @StepContextParameter
+        private transient TaskListener listener;
+
+        @StepContextParameter
+        private transient Run run;
+
+        @Override
+        protected Void run() throws Exception {
+            JiraSite site = JiraSite.get(run.getParent());
+
+            if (site == null) {
+                listener.getLogger().println(Messages.Updater_NoJiraSite());
+                run.setResult(Result.FAILURE);
+            }
+
+            if (StringUtils.isNotEmpty(step.getWorkflowActionName())) {
+                listener.getLogger().println(Messages.JiraIssueUpdateBuilder_UpdatingWithAction(step.getWorkflowActionName()));
+            }
+
+            listener.getLogger().println("[JIRA] JQL: " + step.getJqlSearch());
+
+            try {
+                if (!site.progressMatchingIssues(step.getJqlSearch(), step.workflowActionName, step.getComment(), listener.getLogger())) {
+                    listener.getLogger().println(Messages.JiraIssueUpdateBuilder_SomeIssuesFailed());
+                    run.setResult(Result.UNSTABLE);
+                }
+            } catch (IOException e) {
+                listener.getLogger().println(Messages.JiraIssueUpdateBuilder_Failed());
+                e.printStackTrace(listener.getLogger());
+                run.setResult(Result.FAILURE);
+            }
+
+            return null;
+        }
+    }
+}
--- a/src/main/resources/hudson/plugins/jira/Messages.properties
+++ b/src/main/resources/hudson/plugins/jira/Messages.properties
@@ -37,3 +37,4 @@ IssueSelectorStep.Descriptor.DisplayName=JIRA: Issue selector
 JiraCreateIssueNotifier.DisplayName=JIRA: Create issue
 IssueSelector.ExplicitIssueSelector.DisplayName=Explicit selector
 IssueSelector.JqlIssueSelector.DisplayName=JQL selector
+IssueUpdateStep.Descriptor.DisplayName=JIRA: Progress issues by workflow action
--- a/src/main/resources/hudson/plugins/jira/pipeline/IssueUpdateStep/config.jelly
+++ b/src/main/resources/hudson/plugins/jira/pipeline/IssueUpdateStep/config.jelly
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
+    <f:entry field="jqlSearch" title="JQL Search for selecting issues">
+        <f:textbox />
+    </f:entry>
+    <f:entry field="workflowActionName" title="Target Jira workflow name">
+        <f:textbox />
+    </f:entry>
+    <f:entry field="comment" title="Comment body">
+        <f:textbox />
+    </f:entry>
+</j:jelly>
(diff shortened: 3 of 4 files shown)


oleg-nenashev:
Looks good to me as a code.
But I'm not sure what's the problem with the original JiraIssueUpdateBuilder. It implements SimpleBuildStep, hence it should be available in Pipeline

aatarasoff:
I don't know how builder will look in pipeline, but I think that notation like:
```
jiraIssueUpdate(
  jqlSearch: jql,
  workflowActionName: action,
  comment: "Some comment"
)
```
is really easier for use. Could you provide an example with JiraIssueUpdateBuilder?
And in `pipeline syntax` help I didn't see this builder.

aatarasoff:
Add new pipeline steps. One for updating issue field and one for strict workflow process, cause issue updater set build only Unstable, but not Failure.

src/test/java/hudson/plugins/jira/pipeline/IssueFieldUpdateStepTest.java
```diff
@@ -0,0 +1,123 @@
...
+import static org.hamcrest.collection.IsCollectionWithSize.hasSize;
+import static org.hamcrest.core.IsEqual.equalTo;
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ * Created by aleksandr on 04.07.16.
+ */
+public class IssueFieldUpdateStepTest {
```
warden: Please add class description

src/main/java/hudson/plugins/jira/pipeline/IssueWorkflowActionStep.java
```diff
@@ -0,0 +1,115 @@
...
+import org.jenkinsci.plugins.workflow.steps.AbstractSynchronousNonBlockingStepExecution;
+import org.jenkinsci.plugins.workflow.steps.StepContextParameter;
+import org.kohsuke.stapler.DataBoundConstructor;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Created by aleksandr on 01.07.16.
+ */
+public class IssueWorkflowActionStep  extends AbstractStepImpl {
```
warden: Please add class description

src/main/java/hudson/plugins/jira/pipeline/IssueUpdateStep.java
```diff
@@ -0,0 +1,110 @@
...
+import org.jenkinsci.plugins.workflow.steps.AbstractStepImpl;
+import org.jenkinsci.plugins.workflow.steps.AbstractSynchronousNonBlockingStepExecution;
+import org.jenkinsci.plugins.workflow.steps.StepContextParameter;
+import org.kohsuke.stapler.DataBoundConstructor;
+
+import javax.annotation.Nonnull;
+import java.io.IOException;
+
+/**
+ * Step for updating jira issue with workflow migration
```
warden: Suggest to re-word, as I don't understand it - maybe let's stick to conform to JIRA terminology about workflow *transition* :-)
e.g. "Perform workflow transitions on the issues selected by JQL query"?

src/main/java/hudson/plugins/jira/pipeline/IssueFieldUpdateStep.java
```diff
@@ -0,0 +1,105 @@
...
+import org.jenkinsci.plugins.workflow.steps.AbstractStepImpl;
+import org.jenkinsci.plugins.workflow.steps.AbstractSynchronousNonBlockingStepExecution;
+import org.jenkinsci.plugins.workflow.steps.StepContextParameter;
+import org.kohsuke.stapler.DataBoundConstructor;
+
+import javax.annotation.Nonnull;
+import java.io.IOException;
+
+/**
+ * Step for updating jira issue with workflow migration
```
warden: I think this one is doing something different...

src/main/java/hudson/plugins/jira/pipeline/IssueFieldUpdateStep.java
```diff
@@ -0,0 +1,105 @@
...
+                listener.getLogger().println(Messages.Updater_NoJiraSite());
+                run.setResult(Result.FAILURE);
+            }
+
+            listener.getLogger().println("[JIRA] Update issue with key: " + step.getIssueKey());
+
+            try {
+                if (!site.getSession().updateIssueFieldValue(step.getIssueKey(), step.getFieldName(), step.getFieldValue())) {
+                    listener.getLogger().println(Messages.IssueFieldUpdateStep_Failed());
+                    run.setResult(Result.UNSTABLE);
```
warden: Why UNSTABLE here and FAILURE in other cases? I think we should stick to one failure type in this case.

src/main/java/hudson/plugins/jira/pipeline/IssueUpdateStep.java
```diff
@@ -0,0 +1,110 @@
...
+            if (StringUtils.isNotEmpty(step.getWorkflowActionName())) {
+                listener.getLogger().println(Messages.JiraIssueUpdateBuilder_UpdatingWithAction(step.getWorkflowActionName()));
+            }
+
+            listener.getLogger().println("[JIRA] JQL: " + step.getJqlSearch());
+
+            try {
+                if (!site.progressMatchingIssues(step.getJqlSearch(), step.workflowActionName, step.getComment(), listener.getLogger())) {
+                    listener.getLogger().println(Messages.JiraIssueUpdateBuilder_SomeIssuesFailed());
+                    run.setResult(Result.UNSTABLE);
```
warden: same here

src/main/java/hudson/plugins/jira/pipeline/IssueWorkflowActionStep.java
```diff
@@ -0,0 +1,115 @@
...
+    private final String issueKey;
```
warden: As there is no variable expansion here, I'm wondering what usecase is this going to be, because you'll need to configure "issueKey" as static value in the plugin configuration.
Usually this will not be the case and you'll want to get this from e.g. a variable ?

Maybe you could add variable expansion (see *JiraIssueUpdateBuilder* for example) - that would give some additional great power to those functionalities, what do you think?

warden:
I'll take a look at this tomorrow because it looks like some of the code is duplicated from the original  `JiraIssueUpdateBuilder` class which as @oleg-nenashev pointed out should be usable already as a Pipeline step.
On the other hand, I would warmly welcome the `IssueFieldUpdateStep` functionality as a separate PR!

src/main/java/hudson/plugins/jira/JiraRestService.java
```diff
@@ -18,10 +18,7 @@
 import com.atlassian.jira.rest.client.api.JiraRestClient;
 import com.atlassian.jira.rest.client.api.RestClientException;
 import com.atlassian.jira.rest.client.api.domain.*;
-import com.atlassian.jira.rest.client.api.domain.input.IssueInput;
-import com.atlassian.jira.rest.client.api.domain.input.IssueInputBuilder;
-import com.atlassian.jira.rest.client.api.domain.input.TransitionInput;
-import com.atlassian.jira.rest.client.api.domain.input.VersionInput;
+import com.atlassian.jira.rest.client.api.domain.input.*;
```
artkoshelev: can you please replace wildcard imports?

src/main/java/hudson/plugins/jira/JiraRestService.java
```diff
@@ -368,4 +365,19 @@ public String getBaseApiPath() {
     public Permissions getMyPermissions() throws RestClientException {
         return jiraRestClient.getMyPermissionsRestClient().getMyPermissions(null).claim();
     }
+
+    public boolean updateIssueFieldValue(String issueKey, String fieldName, String fieldValue) {
```
artkoshelev: i think it should be called `updateIssue, since we update any field. and can we use this method inside current `updateIssue` (which is actually updates fix version) and `updateIssueLabels` methods ?

src/main/java/hudson/plugins/jira/JiraRestService.java
```diff
@@ -368,4 +365,19 @@ public String getBaseApiPath() {
...
+
+        try {
+            jiraRestClient.getIssueClient().updateIssue(issueKey, issueInput).get(timeout, TimeUnit.SECONDS);
+            return true;
+        } catch (Exception e) {
+            LOGGER.warning("jira rest client update issue field error. cause: " + e.getMessage());
+        }
+        return false;
+    }
 }
```
artkoshelev: add new line

src/main/resources/hudson/plugins/jira/Messages.properties
```diff
@@ -36,4 +36,9 @@ SearchIssuesStep.Descriptor.DisplayName=JIRA: Search issues
 IssueSelectorStep.Descriptor.DisplayName=JIRA: Issue selector
 JiraCreateIssueNotifier.DisplayName=JIRA: Create issue
 IssueSelector.ExplicitIssueSelector.DisplayName=Explicit selector
-IssueSelector.JqlIssueSelector.DisplayName=JQL selector
\ No newline at end of file
+IssueSelector.JqlIssueSelector.DisplayName=JQL selector
+IssueUpdateStep.Descriptor.DisplayName=JIRA: Progress issues by workflow action
```
artkoshelev: 'Move issues through workflow' ?

src/test/java/hudson/plugins/jira/pipeline/IssueFieldUpdateStepTest.java
```diff
@@ -0,0 +1,123 @@
...
+import org.junit.Before;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.jvnet.hudson.test.JenkinsRule;
+import org.mockito.Mockito;
+import org.mockito.invocation.InvocationOnMock;
+import org.mockito.stubbing.Answer;
+
+import java.io.PrintStream;
+import java.util.*;
```
artkoshelev: Please remove wildcard imports

src/main/resources/hudson/plugins/jira/Messages.properties
warden: Actually progress is a good word, hence i suggest "JIRA: Progress issues through workflow"

src/main/java/hudson/plugins/jira/JiraRestService.java
warden: I'm assuming this is an automatic IDE feature (in IDEA) replacement, that when > 5 imports, wildcards are automatically incurred. Personally this is still debatable for me:
http://stackoverflow.com/questions/147454/why-is-using-a-wild-card-with-a-java-import-statement-bad
Any good reason in Jenkins case about not using wildcard imports that we should maybe put in the README guidelines?

artkoshelev: i just looked through few random classes in jenkins core repo and can't see wildcard imports used. it would be better to add checkstyle plugin to build rather than put it in README

eskobaba:
Please finish work on this so it can be merged, it would enable pipelines for our organization. This is the missing step.

warden:
@eskobaba what about https://github.com/jenkinsci/jira-plugin/pull/101#issuecomment-226165515? did you try this one?

eskobaba:
I need pipeline compatibility to post a value to a customfield
Have i overlooked something?

Get Outlook for iOS<https://aka.ms/o0ukef>

________________________________

rantoniuk:
@aatarasoff are you still using this construct? if so, could you rebase it to the current version?

zbynek:
```
step([$class: 'IssueFieldUpdateStep', 
                    issueSelector: [$class: 'JqlIssueSelector', jql: "id=\"JENKINS-6091\""], fieldId: "10005", fieldValue: "foo" ])
```
already works for updating field with id `10005` (a.k.a. `customfield_10005`)  to `foo`. If the goal is to have the same functionality in Pipeline as in free-style job, all that's need is to add `@Symbol` annotation (https://javadoc.jenkins.io/component/symbol-annotation/org/jenkinsci/Symbol.html)  to the classes that need to be referenced from pipeline. That would allow more concise notation for the example above:

```
jiraIssueFieldUpdate(issueSelector:  jql("id=\"JENKINS-6091\""), fieldId: "10005", fieldValue: "foo" )
```

rantoniuk:
Done in #757

PR #103, opened 2016-09-07
JiraSession.java - "Replace Jira Release" as RegEx

empinator:
Wanting to remove all Versions on an issue based on a regular expression, e.g. "*-SNAPSHOT"

Use Case: Having automated deployments to a test environment, continuously wanting to update the tickets version to the latest snapshot-deployment until QA has confirmed the ticket or if it comes back from QA and gets deployed again. 
This way, only the latest snapshot-version is in the fixVersion list, not affecting the targeted release-version (e.g. "2.5.8-final") next to it.

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/jira-plugin/103)

### DIFF
--- a/src/main/java/hudson/plugins/jira/JiraSession.java
+++ b/src/main/java/hudson/plugins/jira/JiraSession.java
@@ -11,6 +11,8 @@ import java.util.logging.Logger;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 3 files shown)


artkoshelev:
It will brake existing configurations, right?

empinator:
added the possibility to explicitly use RegEx by surrounding the string with slashes: 
e.g.  /.*SNAPSHOT$/

PR #108, opened 2016-11-15
Export the JIRA action so that we can use it from Blue Ocean's REST API

i386:
This change makes it possible to access JIRA actions through the REST API

GET /blue/rest/organizations/jenkins/pipelines/testing/branches/master/runs/2/jira/
```
{
  "_class": "hudson.plugins.jira.JiraBuildAction",
  "issues": [
    {
      "key": "JENKINS-38522",
      "summary": "Long stage names overflow poorly on stage graph"
    },
    {
      "key": "JENKINS-37295",
      "summary": "Redesign the empty states"
    }
  ],
  "serverURL": "https://issues.jenkins-ci.org/"
}
```

Here's whats possible https://github.com/i386/blueocean-jira-plugin

jira

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/jira-plugin/108)

### DIFF
--- a/src/main/java/hudson/plugins/jira/JiraBuildAction.java
+++ b/src/main/java/hudson/plugins/jira/JiraBuildAction.java
@@ -1,30 +1,33 @@
 package hudson.plugins.jira;
 
-import java.util.Arrays;
-import java.util.Collection;
+import java.net.URL;
 import java.util.Set;
-import java.util.SortedSet;
-import java.util.TreeSet;
 
+import com.google.common.base.Objects;
+import com.google.common.collect.Sets;
 import hudson.model.Action;
 import hudson.model.Run;
 import hudson.plugins.jira.model.JiraIssue;
+import org.kohsuke.stapler.export.Exported;
+import org.kohsuke.stapler.export.ExportedBean;
+
+import javax.annotation.Nonnull;
 
 /**
  * JIRA issues related to the build.
  *
  * @author Kohsuke Kawaguchi
  */
+@ExportedBean
 public class JiraBuildAction implements Action {
 
     public final Run<?, ?> owner;
 
-    public JiraIssue[] issues;
+    private Set<JiraIssue> issues;
 
-    public JiraBuildAction(Run<?, ?> owner, Collection<JiraIssue> issues) {
+    public JiraBuildAction(@Nonnull Run<?, ?> owner, @Nonnull Set<JiraIssue> issues) {
         this.owner = owner;
-        this.issues = issues.toArray(new JiraIssue[issues.size()]);
-        Arrays.sort(this.issues);
+        this.issues = Sets.newHashSet(issues);
     }
 
     public String getIconFileName() {
@@ -39,12 +42,26 @@ public class JiraBuildAction implements Action {
         return "jira";
     }
 
+    @Exported(inline = true)
+    public Set<JiraIssue> getIssues() {
+        return issues;
+    }
+
+    @Exported
+    public String getServerURL() {
+        JiraSite jiraSite = JiraSite.get(owner.getParent());
+        URL url = Objects.firstNonNull(jiraSite.alternativeUrl, jiraSite.alternativeUrl);
+        return url != null ? url.toString() : null;
+    }
+
     /**
      * Finds {@link JiraIssue} whose ID matches the given one.
+     * @param issueID e.g. JENKINS-1234
+     * @return JIRAIssue representing the issueID
      */
-    public JiraIssue getIssue(String id) {
+    public JiraIssue getIssue(String issueID) {
         for (JiraIssue issue : issues) {
-            if (issue.id.equals(id)) {
+            if (issue.getKey().equals(issueID)) {
                 return issue;
             }
         }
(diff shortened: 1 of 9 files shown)


i386:
Thanks @warden!

src/main/java/hudson/plugins/jira/JiraBuildAction.java
```diff
@@ -39,23 +42,33 @@ public String getUrlName() {
...
 
+    @Exported(inline = true)
+    public Set<JiraIssue> getIssues() {
+        return issues;
+    }
+
+    @Exported
+    public String getServerURL() {
+        JiraSite jiraSite = JiraSite.get(owner.getParent());
+        URL url = Objects.firstNonNull(jiraSite.alternativeUrl, jiraSite.alternativeUrl);
```
warden: I just noticed this one, was this intentional to have alternativeUrl in both?

i386: Oops that is not correct. There's also a possible NPE here with jiraSite being null. Fixed in https://github.com/jenkinsci/jira-plugin/pull/113

johnou:
Maybe we were a little quick merging this? Unit tests are broken.

johnou:
@i386 would you please create another pull request fixing UpdaterTest?

src/main/java/hudson/plugins/jira/JiraBuildAction.java
```diff
@@ -1,30 +1,33 @@
...
+@ExportedBean
 public class JiraBuildAction implements Action {
 
     public final Run<?, ?> owner;
 
-    public JiraIssue[] issues;
+    private Set<JiraIssue> issues;
 
-    public JiraBuildAction(Run<?, ?> owner, Collection<JiraIssue> issues) {
+    public JiraBuildAction(@Nonnull Run<?, ?> owner, @Nonnull Set<JiraIssue> issues) {
```
johnou: breaks binary compatibility?

src/main/java/hudson/plugins/jira/JiraBuildAction.java
```diff
@@ -39,23 +42,33 @@ public String getUrlName() {
...
         return null;
     }
 
     public void addIssues(Set<JiraIssue> issuesToBeSaved) {
-        SortedSet<JiraIssue> allIssues = new TreeSet<JiraIssue>();
-        allIssues.addAll(issuesToBeSaved);
-        allIssues.addAll(Arrays.asList(this.issues));
-
-        this.issues = allIssues.toArray(new JiraIssue[allIssues.size()]);
+        this.issues.addAll(issuesToBeSaved);
```
johnou: order is no long preserved?

src/main/java/hudson/plugins/jira/Updater.java
```diff
@@ -148,55 +150,51 @@ boolean perform(Run<?, ?> build, TaskListener listener, AbstractIssueSelector se
...
                     continue;
                 }
 
-                logger.println(Messages.FailedToUpdateIssueWithCarryOver(issue.id));
+                logger.println(Messages.FailedToUpdateIssueWithCarryOver(issue.getKey()));
                 logger.println(e.getLocalizedMessage());
             }
-
-            // if no exception is thrown during update, remove from the list as succesfully updated
-            issues.remove(issue);
```
johnou: I believe this is why UpdaterTest now fails.

src/main/java/hudson/plugins/jira/Updater.java
```diff
@@ -148,55 +150,51 @@ boolean perform(Run<?, ?> build, TaskListener listener, AbstractIssueSelector se
...
         }
 
     }
 
-    private static List<JiraIssue> getJiraIssues(Set<String> ids, JiraSession session, PrintStream logger) throws RemoteException {
-        List<JiraIssue> issues = new ArrayList<JiraIssue>(ids.size());
+    private static Set<JiraIssue> getJiraIssues(Set<String> ids, JiraSession session, PrintStream logger) throws RemoteException {
+        Set<JiraIssue> issues = Sets.newHashSet();
```
johnou: why not use ids.size() to initialize size of set?

src/main/java/hudson/plugins/jira/JiraBuildAction.java
i386: if someone is extending this action, yes.

src/main/java/hudson/plugins/jira/JiraBuildAction.java
i386: I wasn't aware that it was significant apart from using `List`. Updating a PR where we use `LinkedHashSet` instead.

src/main/java/hudson/plugins/jira/Updater.java
i386: You are correct. Not sure why I removed these - restored in another PR.

src/main/java/hudson/plugins/jira/Updater.java
i386: Not sure it makes a huge difference given the amount of allocations likely to happen here? Ive restored it anyway.

PR #109, opened 2016-11-18
JENKINS-31164 - Fixed issueTypeId of Bug , so this plugin could work with latest version of JIRA

sam-shen:
fixed issueTypeId of Bug ,so this plugin could work with latest version of JIRA

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/jira-plugin/109)

### DIFF
--- a/src/main/java/hudson/plugins/jira/JiraRestService.java
+++ b/src/main/java/hudson/plugins/jira/JiraRestService.java
@@ -226,7 +226,7 @@ public class JiraRestService {
         IssueInputBuilder builder = new IssueInputBuilder();
         builder.setProjectKey(projectKey)
                 .setDescription(description)
-                .setIssueTypeId(1L) // BUG
+                .setIssueTypeId(10004L) // BUG
                 .setSummary(summary);
 
         if (!assignee.equals(""))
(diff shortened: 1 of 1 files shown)


warden:
This would break previous versions of JIRA which have different BUG ID.
This cannot be a constant value but a value that is populated via REST API from `rest/api/2/issuetype`

sam-shen:
Add getIssueTypeId(String issueTypeName), should work with previous versions of JIRA which have different BUG ID with same "Bug" name.

johnou:
I believe my pull request supersedes this one https://github.com/jenkinsci/jira-plugin/pull/112 👍

src/main/java/hudson/plugins/jira/JiraRestService.java
```diff
@@ -222,11 +223,27 @@ public void releaseVersion(String projectKey, Version version) {
         }
     }
 
+    public Long getIssueTypeId(String issueTypeName) {
+    	List<IssueType> issueTypeList = new ArrayList<IssueType>();
+        Long issueTypeId = 10004L;
+        
+    	issueTypeList = this.getIssueTypes();
+    	for(IssueType issue : issueTypeList) {
+    		LOGGER.warning("issueType: " + issue.getName() + "  "+ issue.getId());
```
johnou: This logging will cause pretty bad spam..

johnou:
Closing in favour of https://github.com/jenkinsci/jira-plugin/pull/112

PR #112, opened 2016-11-23
JENKINS-31164: Configurable issue priority, type and actionIdOnSuccess.

johnou:
Motivation:

When Jenkins creates a JIRA ticket the priority and type ids are hardcoded and may break ticket creation as noted in JENKINS-31164.

Modifications:

Allow the admin to configure issue priority id and type id from the Jenkins job configuration.
Allow the admin to configure actionId on success which allows the plugin to automatically resolve or close tickets.

Result:

Support for JIRA where the type ids are different and greater flexibility when setting up workflows.

This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/jira-plugin/112)

### DIFF
--- a/src/main/java/hudson/plugins/jira/JiraCreateIssueNotifier.java
+++ b/src/main/java/hudson/plugins/jira/JiraCreateIssueNotifier.java
@@ -1,15 +1,13 @@
 package hudson.plugins.jira;
 
-import com.atlassian.jira.rest.client.api.domain.BasicComponent;
-import com.atlassian.jira.rest.client.api.domain.Component;
 import com.atlassian.jira.rest.client.api.domain.Issue;
+import com.atlassian.jira.rest.client.api.domain.IssueType;
+import com.atlassian.jira.rest.client.api.domain.Priority;
 import com.atlassian.jira.rest.client.api.domain.Status;
 import com.google.common.base.Splitter;
-import com.google.common.collect.Lists;
 import hudson.EnvVars;
 import hudson.Extension;
 import hudson.Launcher;
-import hudson.Util;
 import hudson.model.AbstractBuild;
 import hudson.model.AbstractProject;
 import hudson.model.BuildListener;
@@ -20,6 +18,7 @@ import hudson.tasks.BuildStepMonitor;
 import hudson.tasks.Notifier;
 import hudson.tasks.Publisher;
 import hudson.util.FormValidation;
+import hudson.util.ListBoxModel;
 import net.sf.json.JSONObject;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.QueryParameter;
@@ -31,10 +30,9 @@ import java.io.FileNotFoundException;
 import java.io.FileReader;
 import java.io.IOException;
 import java.io.PrintWriter;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
 import java.util.logging.Logger;
+
+import static hudson.plugins.jira.JiraRestService.BUG_ISSUE_TYPE_ID;
 
 /**
  * When a build fails it creates jira issues.
@@ -50,6 +48,9 @@ public class JiraCreateIssueNotifier extends Notifier {
     private String testDescription;
     private String assignee;
     private String component;
+    private Long typeId;
+    private Long priorityId;
+    private Integer actionIdOnSuccess;
 
     enum finishedStatuses {
         Closed,
@@ -58,13 +59,22 @@ public class JiraCreateIssueNotifier extends Notifier {
     }
 
     @DataBoundConstructor
-    public JiraCreateIssueNotifier(String projectKey, String testDescription, String assignee, String component) {
+    public JiraCreateIssueNotifier(String projectKey, String testDescription, String assignee, String component, Long typeId,
+                                   Long priorityId, Integer actionIdOnSuccess) {
         if (projectKey == null) throw new IllegalArgumentException("Project key cannot be null");
         this.projectKey = projectKey;
 
         this.testDescription = testDescription;
         this.assignee = assignee;
         this.component = component;
+        this.typeId = typeId;
+        this.priorityId = priorityId;
+        this.actionIdOnSuccess = actionIdOnSuccess;
+    }
+
+    @Deprecated
+    public JiraCreateIssueNotifier(String projectKey, String testDescription, String assignee, String component) {
+        this(projectKey, testDescription, assignee, component, null, null, null);
     }
 
     public String getProjectKey() {
@@ -97,6 +107,18 @@ public class JiraCreateIssueNotifier extends Notifier {
 
     public void setComponent(String component) {
         this.component = component;
+    }
+
+    public Long getTypeId() {
+        return typeId;
+    }
+
+    public Long getPriorityId() {
+        return priorityId;
+    }
+
+    public Integer getActionIdOnSuccess() {
+        return actionIdOnSuccess;
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


johnou:
@oleg-nenashev could I trouble you for a review?

johnou:
@warden PTAL

artkoshelev:
Please describe the purpose of your changes

johnou:
@artkoshelev updated, please let me know if you would like me to explain in greater detail.

johnou:
Tests failed due to https://github.com/jenkinsci/jira-plugin/pull/108

johnou:
@i386 could I get your feedback on this one please?

i386:
@johnou left a comment about using the internal JIRA ids. They are really difficult to figure out unless you go through the source code of pages.

johnou:
@i386 how about string type and try to fallback to ids if the input looks like a number? or is that too "magic"?

i386:
@johnou maybe a bit too magic and I am not sure why you would want that fallback behaviour?

johnou:
@warden @artkoshelev gentle ping, could I collect your thoughts too?

i386:
@johnou opt for whatever means users have to think or do less

johnou:
Our TechOps team had a vote and the consensus was that supporting both would be the ideal fit for all, we would opt for id as the name is more volatile in our projects but this may not be the case for others where name might make more sense.

johnou:
@warden finally ready for review!

sam-shen:
Why don't you add statusId together? In my case, different typeId will have different statusId of "Done/Closed"

johnou:
@sam-shen you could just configure "Bug" or whatever you want for Issue type. I am not sure what Done / Closed has to do with the status id?

src/main/java/hudson/plugins/jira/JiraCreateIssueNotifier.java
```diff
@@ -50,6 +50,9 @@
     private String testDescription;
     private String assignee;
     private String component;
+    private String priority;
+    private String type;
+    private Integer actionIdOnSuccess;
 
     enum finishedStatuses {
```
sam-shen: Could "finishedStatuses" have a dropdown list instand of hardcode here ?

johnou: Drop down list, probably not, maybe override in csv of finished statuses? These are checked by taking the string value of status and then equals ignore case against the enum. Unfortunately I don't have the development time to implement it at this stage and I think it would be better suited in a follow up pull request anyway.

src/main/java/hudson/plugins/jira/JiraCreateIssueNotifier.java
```diff
@@ -165,12 +194,54 @@ private Issue createJiraIssue(AbstractBuild<?, ?> build, String filename) throws
...
+            for (IssueType t : session.getIssueTypes()) {
+                if (type.equalsIgnoreCase(t.getName())) {
+                    return t.getId();
+                }
+            }
+            if (isNumber(type)) {
+                return Long.valueOf(type.trim());
+            }
+        }
+        return JiraRestService.BUG_ISSUE_TYPE_ID;
```
warden: Let's add an INFO logger here that we're returning default  value.

src/main/java/hudson/plugins/jira/JiraCreateIssueNotifier.java
```diff
@@ -165,12 +194,54 @@ private Issue createJiraIssue(AbstractBuild<?, ?> build, String filename) throws
...
+                }
+            }
+            if (isNumber(priority)) {
+                return Long.valueOf(priority.trim());
+            }
+        }
+        return null;
+    }
+
+    private final Pattern pattern = Pattern.compile("-?\\d+");
```
warden: to the top and more meaningful variable name :-)
(and maybe the actual regex in a constant?)

src/main/java/hudson/plugins/jira/JiraCreateIssueNotifier.java
```diff
@@ -190,14 +261,15 @@ private Status getStatus(AbstractBuild<?, ?> build, String id) throws IOExceptio
...
      * @param id
      * @param comment
      * @throws IOException
      */
-    private void addComment(AbstractBuild<?, ?> build, String id, String comment) throws IOException {
-
+    private void addComment(AbstractBuild<?, ?> build, BuildListener listener, String id, String comment) throws IOException {
         JiraSession session = getJiraSession(build);
         session.addCommentWithoutConstrains(id, comment);
+        listener.getLogger().println("Commented JIRA issue " + id);
```
warden: [JIRA] Commented issue

src/main/resources/hudson/plugins/jira/JiraCreateIssueNotifier/help-type.html
```diff
@@ -0,0 +1,4 @@
+<div>
+    <strong>Optional</strong> <br/>
+    This will be the type of the JIRA issue, for project default leave empty.
```
warden: In the code I see either name (Bug) is permitted or a numeric value, but this help message doesn't mention this..

johnou:
@warden PTAL

johnou:
@warden I'll push better UI changes soon.

johnou:
jiraplugin

johnou:
Unfortunately I had to leave Action Id On Success as an id field for now as there is no API endpoint I could find to return workflows.

johnou:
@warden coverage should be fixed now.

johnou:
Still 0%, weird.. at least the coverage report on my local machine has hits now.

coverage

johnou:
If I have further ideas on what might be wrong I will create a follow up pull request.

warden:
Please rebase, coverage should be resurrected... :)

johnou:
@warden rebased!

i386:
@johnou I think this would be better if you could select the JIRA server as a drop down list then only show the items for the selected JIRA server in the other fields. This will reduce the length of the select values. As the server URL gets longer it becomes harder to read the issue type and priority labels. 

I would also change the fields so the format was `Task (10010)` etc

src/main/resources/hudson/plugins/jira/JiraCreateIssueNotifier/config.jelly
```diff
@@ -1,14 +1,23 @@
...
-  <f:entry title="${%Assignee}" field="assignee">
-            <f:textbox/>
-      </f:entry>
-   <f:entry title="${%Description Of Test}" field="testDescription">
-          <f:textarea/>
+    </f:entry>
+    <f:entry title="${%Assignee}" field="assignee">
+        <f:textbox/>
+    </f:entry>
+    <f:entry title="${%Description Of Test}" field="testDescription">
```
i386: `Description of Test` rather than `Description Of Test`

johnou:
@i386 it's actually tied to the build in some cases and I'm not really sure how to pull that up into the descriptor..

```java
    /**
     * Gets the effective {@link JiraSite} associated with the given project.
     *
     * @return null
     *         if no such was found.
     */
    public static JiraSite get(Job<?, ?> p) {
        JiraProjectProperty jpp = p.getProperty(JiraProjectProperty.class);
        if (jpp != null) {
            JiraSite site = jpp.getSite();
            if (site != null) {
                return site;
            }
        }

        // none is explicitly configured. try the default ---
        // if only one is configured, that must be it.
        JiraSite[] sites = JiraProjectProperty.DESCRIPTOR.getSites();
        if (sites.length == 1) {
            return sites[0];
        }

        return null;
    }
```

johnou:
@warden thanks!

johnou:
@warden I think we may have found a small bug in the way the plugin handles the ticket status, if you can wait a day or two before pushing out a new version of the plugin I will submit a pull request.

johnou:
https://github.com/jenkinsci/jira-plugin/pull/117

PR #113, opened 2016-11-24
Fix calculation of server URL for JiraBuildAction

i386:
PTAL @warden

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/jira-plugin/113)


artkoshelev:
Can you please describe problem you solve and provide test reproducing it?

PR #114, opened 2016-11-24
Add support for parsing branch names containing JIRA IDs

i386:
Add JiraJobAction to store a reference a JIRAIssue if a MultibranchPipeline has a branch with a issue key in its name.

It is a common pattern to include a JIRA ticket reference in a branch name like `feature/JENKINS-1234`

This action will allow us to link to it like this:
blueocean 477 2016-11-24 20-29-53

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/jira-plugin/114)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -286,6 +286,16 @@
 	  <optional>true</optional>
 	</dependency>
     <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>branch-api</artifactId>
+      <version>1.0</version>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-job</artifactId>
+      <version>${workflow.version}</version>
+    </dependency>
+    <dependency>
         <groupId>org.jvnet.hudson.plugins</groupId>
         <artifactId>perforce</artifactId>
         <version>1.3.9</version>
@@ -338,12 +348,6 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <artifactId>workflow-scm-step</artifactId>
-      <version>${workflow.version}</version>
-      <scope>test</scope>
-    </dependency>
-   	<dependency>
-      <groupId>org.jenkins-ci.plugins.workflow</groupId>
-      <artifactId>workflow-job</artifactId>
       <version>${workflow.version}</version>
       <scope>test</scope>
     </dependency>
--- a/src/main/java/hudson/plugins/jira/JiraSite.java
+++ b/src/main/java/hudson/plugins/jira/JiraSite.java
@@ -71,17 +71,13 @@ public class JiraSite extends AbstractDescribableImpl<JiraSite> {
     /**
      * URL of JIRA for Jenkins access, like <tt>http://jira.codehaus.org/</tt>.
      * Mandatory. Normalized to end with '/'
-     * @deprecated use @{getUrl}
      */
-    @Deprecated
     public final URL url;
 
     /**
      * URL of JIRA for normal access, like <tt>http://jira.codehaus.org/</tt>.
      * Mandatory. Normalized to end with '/'
-     * @deprecated use @{getUrl}
      */
-    @Deprecated
     public final URL alternativeUrl;
 
     /**
--- a/src/main/java/hudson/plugins/jira/JiraSite.java
+++ b/src/main/java/hudson/plugins/jira/JiraSite.java
@@ -5,6 +5,8 @@ import com.atlassian.jira.rest.client.api.RestClientException;
 import com.atlassian.jira.rest.client.api.domain.Issue;
 import com.atlassian.jira.rest.client.api.domain.Version;
 import com.atlassian.jira.rest.client.internal.async.AsynchronousJiraRestClientFactory;
+import com.google.common.base.*;
+import com.google.common.base.Objects;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import hudson.Extension;
@@ -69,13 +71,17 @@ public class JiraSite extends AbstractDescribableImpl<JiraSite> {
     /**
      * URL of JIRA for Jenkins access, like <tt>http://jira.codehaus.org/</tt>.
      * Mandatory. Normalized to end with '/'
+     * @deprecated use @{getUrl}
      */
+    @Deprecated
     public final URL url;
 
     /**
      * URL of JIRA for normal access, like <tt>http://jira.codehaus.org/</tt>.
      * Mandatory. Normalized to end with '/'
+     * @deprecated use @{getUrl}
      */
+    @Deprecated
     public final URL alternativeUrl;
 
     /**
@@ -280,6 +286,14 @@ public class JiraSite extends AbstractDescribableImpl<JiraSite> {
                 .createWithBasicHttpAuthentication(uri, userName, password.getPlainText());
         int usedTimeout = timeout != null ? timeout : JiraSite.DEFAULT_TIMEOUT;
         return new JiraSession(this, new JiraRestService(uri, jiraRestClient, userName, password.getPlainText(), usedTimeout));
+    }
+
+    /**
+     * @return the server URL
+     */
+    @Nullable
+    public URL getUrl() {
+        return Objects.firstNonNull(this.url, this.alternativeUrl);
     }
 
     /**
--- a/src/main/java/hudson/plugins/jira/JiraBuildAction.java
+++ b/src/main/java/hudson/plugins/jira/JiraBuildAction.java
@@ -50,7 +50,7 @@ public class JiraBuildAction implements Action {
     @Exported
     public String getServerURL() {
         JiraSite jiraSite = JiraSite.get(owner.getParent());
-        URL url = jiraSite != null ? Objects.firstNonNull(jiraSite.url, jiraSite.alternativeUrl) : null;
+        URL url = jiraSite != null ? jiraSite.getUrl() : null;
         return url != null ? url.toString() : null;
     }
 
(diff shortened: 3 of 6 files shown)


src/test/java/hudson/plugins/jira/pipeline/JiraJobActionTest.java
```diff
@@ -0,0 +1,43 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+public class JiraJobActionTest {
+
+    @Test
+    public void testDetectBranchNameIssue() throws Exception {
+        JiraSite site = mock(JiraSite.class);
+        WorkflowJob job = mock(WorkflowJob.class);
```
i386: @jglick any idea on how to mock a WorkflowJob? Mockito won't mock finals and I don't think setting up a full multibranch pipeline backed by a git repo, scanning, etc would be appropriate here.

johnou: Could you mock the interface Job instead?

i386: @johnou would fail the type check within the `JiraJobAction` :/

johnou: @i386 can you use interfaces to check the type of job? If so you could use something like this..

```java
Job job = mock(Job.class,
    withSettings().extraInterfaces(WorkflowJobInterface.class));
```

johnou: and if that doesn't work, you could always extract the check to a protected method and override that in the test but maybe @jglick has a trick or two up his sleeve.

i386: Actually I can. Updated.

i386:
This can be reviewed now

johnou:
Tests failed due to regression in hudson.plugins.jira.UpdaterTest.testIssueIsRemovedFromCarryOverListAfterSubmission which is fixed with https://github.com/jenkinsci/jira-plugin/pull/115

pom.xml
```diff
@@ -280,6 +280,16 @@
 	  <optional>true</optional>
 	</dependency>
     <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>branch-api</artifactId>
+      <version>1.0</version>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
```
johnou: duplicate declaration of workflow-job? test scope was removed below.

src/main/java/hudson/plugins/jira/pipeline/JiraJobAction.java
```diff
@@ -0,0 +1,124 @@
...
+        this.jiraIssue = jiraIssue;
+    }
+
+    @Exported
+    public JiraIssue getJiraIssue() {
+        return jiraIssue;
+    }
+
+    @Exported
+    public String getServerURL() {
```
johnou: nullable annotation?

src/main/java/hudson/plugins/jira/pipeline/JiraJobAction.java
```diff
@@ -0,0 +1,124 @@
...
+    /**
+     * Adds a {@link JiraJobAction} to a {@link WorkflowJob} if it belongs to a {@link MultiBranchProject}
+     * and its name contains an JIRA issue key
+     * @param job to add the property to
+     * @param site to fetch issue data
+     * @throws IOException if something goes wrong fetching the JIRA issue
+     */
+    public static void setAction(@Nonnull Job<?, ?> job, @Nonnull JiraSite site) throws IOException {
+        // If there is already a action set then skip
+        if (job.getAction(JiraJobAction.class) != null) {
```
johnou: is it a misconfiguration? should we log warn message?

src/main/java/hudson/plugins/jira/pipeline/JiraJobAction.java
```diff
@@ -0,0 +1,124 @@
...
+            return;
+        }
+
+        // Exclude all non-multibranch workflow jobs
+        if (!(job.getParent() instanceof MultiBranchProject)) {
+            return;
```
johnou: in what cases would this be invoked with a non MultiBranchProject? is it a misconfiguration? should we log warn message?

src/main/java/hudson/plugins/jira/pipeline/JiraJobAction.java
```diff
@@ -0,0 +1,124 @@
...
+        }
+
+        // Find the first JIRA issue key in the branch name
+        // If it exists, create the action and set it
+        String issueKey = null;
```
johnou: extract into method? could we unit test this too?

src/test/java/hudson/plugins/jira/pipeline/JiraJobActionTest.java
```diff
@@ -0,0 +1,82 @@
...
+    @Mock
+    JiraSite site;
+
+    @Mock
+    Job job;
+
+    @Mock
+    MultiBranchProject mbp;
+
+    final JiraIssue issue = new JiraIssue("EXAMPLE-123", "I like cake");
```
johnou: me too.

src/main/java/hudson/plugins/jira/pipeline/JiraJobAction.java
i386: Fixed

pom.xml
i386: Fixed

src/main/java/hudson/plugins/jira/pipeline/JiraJobAction.java
i386: When it is a plain workflow job I believe (there are two types)

src/test/java/hudson/plugins/jira/pipeline/JiraJobActionTest.java
i386: 🍰 🍰 🍰

src/main/java/hudson/plugins/jira/pipeline/JiraJobAction.java
i386: Its covered reasonably well in the tests already.

src/main/java/hudson/plugins/jira/pipeline/JiraJobAction.java
i386: `setAction` is called on every run. If the fetch to JIRA fails, then for the next run we have the opportunity to fetch again. Eventually the data is stored on the action and this branch is executed.

src/main/java/hudson/plugins/jira/pipeline/JiraJobAction.java
johnou: ok so this might be invoked in normal circumstances and logging info / warn is just spam.

src/main/java/hudson/plugins/jira/pipeline/JiraJobAction.java
johnou: I think extracting it into its own private method would improve readability, huge fan of short concise methods which have one job.

i386: I used to be the same and was discouraged - stylistic thing ;)

warden:
OK, so this is about the discussion of using commit message JIRA numbers instead of branch_name_suffixes. :-)
I'm not against this one, but what about supporting two JIRA tickets in one branch name? :)

i386:
@warden that's not how this practice works. It's supposed to be a 1to1 mapping.

src/main/java/hudson/plugins/jira/pipeline/JiraJobAction.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    /**
+     * @return url of the JIRA server
+     */
+    @Exported
+    @Nullable
+    public String getServerURL() {
+        JiraSite jiraSite = JiraSite.get(owner);
+        URL url = jiraSite != null ? Objects.firstNonNull(jiraSite.url, jiraSite.alternativeUrl) : null;
```
warden: mmm, how about moving this to JiraSite actually?

src/main/java/hudson/plugins/jira/pipeline/JiraJobAction.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    @Override
+    public String getIconFileName() {
+        return null;
+    }
+
+    @Override
+    public String getDisplayName() {
+        return "jira";
```
warden: JIRA

i386:
@warden pushed the feedback you have requested

warden:
Thanks! 
Umm, where did the coverage go ?! Let me investigate..

warden:
Please rebase, coverage should be resurrected... :)

i386:
@warden I've rebased with master and squashed my commits. Does this look good to you?

src/main/java/hudson/plugins/jira/JiraSite.java
```diff
@@ -69,13 +71,17 @@
...
      */
+    @Deprecated
     public final URL url;
 
     /**
      * URL of JIRA for normal access, like <tt>http://jira.codehaus.org/</tt>.
      * Mandatory. Normalized to end with '/'
+     * @deprecated use @{getUrl}
      */
+    @Deprecated
```
warden: This is actually not Deprecated, see JiraChangeLogAnnotator :-)
(this is not obvious there I know... needs refactoring - alternativeUrl is used for links generated in HTML output, but not for Jenkins communication)
I'd say -> use getAlternativeUrl().
We should assume that getUrl always != null but alternativeUrl = getUrl if null

warden:
LGTM only one small comment :-)
@artkoshelev @olamy any comments from you guys?

i386:
@warden Ill remove those `Deprecated` annotations for now. There needs to be a refactor to remove them that is bigger than the scope of this PR.

PR #115, opened 2016-11-24
Fix calculation of server URL for JiraBuildAction

i386:
@johnou this contains the feedback you requested + fixed tests from #108 

Apologies for the breakages.

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/jira-plugin/115)

### DIFF
--- a/src/test/java/hudson/plugins/jira/UpdaterTest.java
+++ b/src/test/java/hudson/plugins/jira/UpdaterTest.java
@@ -162,7 +162,7 @@ public class UpdaterTest {
 
         this.updater = new Updater(build2.getProject().getScm());        
         
-        final List<JiraIssue> ids = Lists.newArrayList(new JiraIssue("FOOBAR-1", null), new JiraIssue("FOOBAR-2", null));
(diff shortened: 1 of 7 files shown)


src/test/java/hudson/plugins/jira/UpdaterTest.java
```diff
@@ -297,7 +297,7 @@ public Object answer(InvocationOnMock invocation) throws Throwable {
         );
 
         // expected issue list
-        final List<JiraIssue> expectedIssuesToCarryOver = new ArrayList<JiraIssue>();
+        final Set<JiraIssue> expectedIssuesToCarryOver = Sets.newHashSet();
```
johnou: Sets.newLinkedHashSet(); ? If running JDK 8+ order as stated in the javadocs is not predictable and can cause random test assertion failures.

johnou: Might want to use it in other places too if ordering was predictable.

johnou:
@i386 thanks!

i386:
@johnou I went through and 's/HashSet/LinkedHashSet/' most of the tests. At least now it is consistent.

johnou:
lgtm!

PR #118, opened 2016-12-13
[WIP] Option to disable jira annotating the changelogs.

christ66:
@reviewbybees

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/jira-plugin/118)

### DIFF
--- a/src/main/java/hudson/plugins/jira/JiraChangeLogAnnotator.java
+++ b/src/main/java/hudson/plugins/jira/JiraChangeLogAnnotator.java
@@ -45,6 +45,11 @@ public class JiraChangeLogAnnotator extends ChangeLogAnnotator {
             return;    // not configured with JIRA
         }
 
+        if (site.getDisableChangelogAnnotations()) {
+            LOGGER.finest("Jira annotations disabled for changelogs. Skipping.");
+            return;
+        }
+
         LOGGER.log(Level.FINE, "Using site: {0}", site.getUrl());
 
         // if there's any recorded detail information, try to use that, too.
--- a/src/main/java/hudson/plugins/jira/JiraSite.java
+++ b/src/main/java/hudson/plugins/jira/JiraSite.java
@@ -118,6 +118,13 @@ public class JiraSite extends AbstractDescribableImpl<JiraSite> {
     public final boolean recordScmChanges;
 
     /**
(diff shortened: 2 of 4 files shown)


christ66:
To add some more context to this:

Sometimes when viewing an individual build, or the changeset of a project it might take a very long time because the changeset for one build might be very large. I encountered a performance issue when the build had several hundred in the changeset (was not built in a very long time).

Snippet of thread dump causing issue:

```
at com.google.common.cache.LocalCache$LocalManualCache.get(LocalCache.java:4764)
at hudson.plugins.jira.JiraSite.getIssue(JiraSite.java:424)
at hudson.plugins.jira.JiraChangeLogAnnotator.annotate(JiraChangeLogAnnotator.java:101)
at hudson.scm.ChangeLogSet$Entry.getMsgAnnotated(ChangeLogSet.java:252)
at sun.reflect.GeneratedMethodAccessor611.invoke(Unknown Source)
```

It might also be that the performance issue I am encountering is because I am hitting the max api calls, which caused some of the queries to hang. From looking at the `JiraSite.getIssue` it looks like that this will make sure that the jira exists, however it could be that there's a rate limit for the number of Rest API calls I can make.

warden:
Do you think it's a good idea to disable it on a "site" level?
This seems like a corner-case to me and I think it might be better to actually disable it in specific project/job rather than on the whole site.

stephenc:
I think it is a good idea to be able to do both: per project and entire site

christ66:
I will update the pr to provide an option for project level as well.

warden:
@christ66 will you have a chance to take a look on this by the end of Jan?

warden:
ping @christ66

artkoshelev:
Can you add test to document new behavior?

warden:
As I don't see any friction in here, I'll merge it as is. 
@christ66 feel free to submit a new PR with folder level if you have some spare time.

warden:
PR with tests waiting for you @christ66 .
When writing the tests I realized actually that this is not the best solution to disable annotation completely on the site level as this would disable also the "Linking JIRA issues to Run" feature for all builds on this JIRA site.

I'm thinking it would be better to either:
* add option to disable this on a specific Job level
* or introduce a changelog limit (e.g. 20) when parsing the changelog

PR #119, opened 2017-01-10
Add JiraIssuesSizeEnvironmentVariableBuilder build step

mupakoz:
Adds JiraIssuesSizeEnvironmentVariableBuilder build step which enables saving number of issues matching a JQL query in a environment variable of specified name.

I came up with this idea for a new feature when I wanted to fail the build in case there are issues on JIRA that were resolved but still untested. Thanks to the plugin we can save the number of issues we are interested in and then use it in the next steps (for example a shell script or a conditional step).

Please let me know what you think!


warden:
How about extending JiraEnvironmentContributingAction + JiraEnvironmentVariableBuilder to provide issue number in ENV from the used AbstractIssueSelector?

mupakoz:
It seems that there is a problem with the JiraEnvironmentVariableBuilder if I understand the files correctly (forgive me I am a Jenkins plugin developer newbie). In the config.jelly there is a condition for the issue selector: 'if descriptor.hasIssueSelectors()' but the method 'hasIssueSelectors' is not defined in the class DescriptorImpl. That's why the control for choosing the selector does not show up. I can fix the error in my PR.

Do you think the ENV var name should be set arbitrary or the user should be able to set it?

mupakoz:
I just thought that when using JiraEnvironmentVariableBuilder with arbitrary variable names we would be limited to only one set of issues per job.

warden:
1. see https://github.com/jenkinsci/jira-plugin/blob/master/src/main/java/hudson/plugins/jira/JiraEnvironmentContributingAction.java#L14
I would use JIRA_ISSUES_SIZE

2. Yes, the argument for using it multiple times is a valid one. I'm curious what's your use case for this?

mupakoz:
1. My proposal is to use JIRA_ISSUES_SIZE by default but to put a text input in the 'advanced' section: so it would be configurable.
2. In my use case I only need one number, but I can imagine a scenario where two lists of issues would be needed, but we could leave that until a real feature request will come.

warden:
1. agreed, if you want to code it then I'm fine :)
2. 👍

src/main/java/hudson/plugins/jira/JiraIssuesSizeEnvironmentVariableBuilder.java
```diff
@@ -0,0 +1,141 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package hudson.plugins.jira;
+
+import hudson.*;
```
warden: We have agreed approach not to use wildcard imports, could you expand those?

src/main/java/hudson/plugins/jira/JiraIssuesSizeEnvironmentContributingAction.java
```diff
@@ -0,0 +1,38 @@
...
+        return variableName;
+    }
+
+    public JiraIssuesSizeEnvironmentContributingAction(int issuesNumber, String variableName) {
+        this.issuesNumber = issuesNumber;
+        this.variableName = variableName;
+    }
+    
+    @Override
+    public void buildEnvVars(AbstractBuild<?, ?> ab, EnvVars ev) {
```
warden: no two letter variable names please :)

warden:
@mupakoz ping, are you able to address review comments?

mupakoz:
Yes, sorry for my inactivity. I will fix the implementation and the comments by the end of this week.

warden:
@mupakoz ping? :-)

mupakoz:
@warden done. Let me know what you think.

mupakoz:
I also added the hasIssueSelectors() method because I think it was missing in the class.

src/main/resources/hudson/plugins/jira/JiraEnvironmentVariableBuilder/config.jelly
```diff
@@ -1,5 +1,10 @@
 <j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
     <j:if test="${descriptor.hasIssueSelectors()}">
         <f:dropdownDescriptorSelector field="issueSelector" title="${%Issue selector}"/>
-    </j:if>  
+    </j:if>
+    <f:advanced>
+        <f:entry title="${%Issues size variable name}" field="issuesSizeVariableName">
+            <f:textbox default="JIRA_ISSUES_SIZE"/>
```
warden: better to use word COUNT instead of SIZE.
size is usually used when creating an object of specific size to show how many elements can fit in collection, not how many there are

warden: Applies to all files of course :)

rantoniuk:
Follow-up in https://github.com/jenkinsci/jira-plugin/pull/707.

PR #120, opened 2017-02-03
[JENKINS-41690] Add setSites() implementation which accepts List<JiraSite>

stevenbaker:
Currently, JiraProjectProperty's `setSites()` method simply takes a single JiraSite object and adds it to the instance.  In practice, therefore, it behaves as more of an `addSite()` method than a `setSites()` method.  

This can be problematic if you're scripting out configuration of this plugin as part of something like an `init.groovy.d` script.  In that scenario, every time Jenkins starts up, you'll end up with another JIRA site on your instance.

To fix this, going forward, `setSites()` should include an additional implementation that:

* Takes a list of JiraSite objects (`List<JiraSite>`)
* Then completely replaces JiraProjectProperty.sites with what is in your list

This would allow scripted configuration of the JIRA plugin to be idempotent.

Details here:
https://issues.jenkins-ci.org/browse/JENKINS-41690

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/jira-plugin/120)

### DIFF
--- a/src/main/java/hudson/plugins/jira/JiraProjectProperty.java
+++ b/src/main/java/hudson/plugins/jira/JiraProjectProperty.java
@@ -92,8 +92,12 @@ public class JiraProjectProperty extends JobProperty<Job<?, ?>> {
         }
 
         public void setSites(JiraSite site) {
-            sites.clear();
             sites.add(site);
+        }
+
+        public void setSites(List<JiraSite> jiraSites) {
+            sites.clear();
+            sites.addAll(jiraSites);
         }
 
         public JiraSite[] getSites() {
--- a/src/main/java/hudson/plugins/jira/JiraProjectProperty.java
+++ b/src/main/java/hudson/plugins/jira/JiraProjectProperty.java
@@ -92,6 +92,7 @@ public class JiraProjectProperty extends JobProperty<Job<?, ?>> {
         }
 
         public void setSites(JiraSite site) {
+            sites.clear();
             sites.add(site);
         }
 
--- a/src/main/java/hudson/plugins/jira/JiraProjectProperty.java
+++ b/src/main/java/hudson/plugins/jira/JiraProjectProperty.java
@@ -15,6 +15,7 @@ import org.kohsuke.stapler.StaplerRequest;
 
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.util.List;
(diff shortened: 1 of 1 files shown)


warden:
You're assuming that Job can be linked with only one JiraSite, but that's not true, see:
https://github.com/jenkinsci/jira-plugin/blob/master/src/main/java/hudson/plugins/jira/JiraCreateIssueNotifier.java#L455
https://github.com/jenkinsci/jira-plugin/blob/master/src/main/java/hudson/plugins/jira/JiraProjectProperty.java#L59

stevenbaker:
@warden You're correct.  I did make that assumption, based on the fact that "setSites()" only takes a single JiraSite object, and not a collection of them:  

```
public void setSites(JiraSite site) {
    sites.add(site);
}
```

Thanks for that feedback.  :-)  

How about this - perhaps what should happen is something more like this:

```
public void setSites(List<JiraSite> jiraSites) {
    sites.clear();
    sites.addAll(jiraSites);
}
```

That way, if you have one JiraSite, or as you pointed out if you have many of them, you can know that if you call setSites() with "n" sites in the list, then that's exactly how many that will end up in the project.

Turns out, this is exactly what's already happening over in the Confluence plugin:

https://github.com/jenkinsci/confluence-publisher-plugin/blob/master/src/main/java/com/myyearbook/hudson/plugins/confluence/ConfluencePublisher.java#L736-L739

What they do there is:
```
public void setSites(List<ConfluenceSite> sites) {
    this.sites.clear();
    this.sites.addAll(sites);
}
```

I'll revise my PR to do something more like this.

stevenbaker:
Looks like I've got a build failure.  I'll take a look.

stevenbaker:
Build green again.  First time contributor to this plugin, so I'm definitely curious if this one makes sense to y'all or if my thinking is off.

warden:
Sure, thanks for the contribution! 
I'm okay with this, but I think we should either refactor and remove the setSites(JiraSite) or at least mark it as @Deprecated. Also remember to squash your commits in the end :)

warden:
@stevenbaker will you have a chance to work on this yet?

stevenbaker:
The general issue I was trying to solve in this PR (being able to set a List of JiraSites) was fixed in #168.  Closing this one out.

PR #123, opened 2017-04-12
[JENKINS-33222] Better support for parallel builds

elliot-nelson:
### GOAL

* Fix [JENKINS-33222](https://issues.jenkins-ci.org/browse/JENKINS-33222)

On our Jenkins instance, we have many jobs that build 3-4 builds in parallel, often with variable run times.  It's not uncommon for failed builds to end in just a few minutes, while lower-number builds are still running.

The JIRA plugin breaks this type of setup, because it blocks while waiting for the prior build to complete before it will report.  We'd like to use the JIRA plugin, but without blocking on prior builds.

### CHANGES

* Change all references to `BuildStepMonitor.BUILD` to `BuildStepMonitor.NONE` (telling Jenkins we do not require the previous build to complete before running any of the plugin's Builders).

* Change all references to `getPreviousBuild()` to `getPreviousCompletedBuild()`.  By getting the last completed build, we're allowing the way the plugin builds history in JIRA to reflect the order in which builds completed, instead of the order they started.

### TESTS

Tests pass locally, and a local cut of the plugin is now behaving correctly (the JIRA plugin can report issues immediately even if the previous build is still executing).

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/jira-plugin/123)

### DIFF
--- a/src/main/java/hudson/plugins/jira/JiraCreateIssueNotifier.java
+++ b/src/main/java/hudson/plugins/jira/JiraCreateIssueNotifier.java
@@ -130,7 +130,7 @@ public class JiraCreateIssueNotifier extends Notifier {
     public static final DescriptorImpl DESCRIPTOR = new DescriptorImpl();
 
     public BuildStepMonitor getRequiredMonitorService() {
-        return BuildStepMonitor.BUILD;
+        return BuildStepMonitor.NONE;
     }
 
     @Override
@@ -144,7 +144,7 @@ public class JiraCreateIssueNotifier extends Notifier {
         Result currentBuildResult = build.getResult();
 
         Result previousBuildResult = null;
-        AbstractBuild<?, ?> previousBuild = build.getPreviousBuild();
+        AbstractBuild<?, ?> previousBuild = build.getPreviousCompletedBuild();
 
         if (previousBuild != null) {
             previousBuildResult = previousBuild.getResult();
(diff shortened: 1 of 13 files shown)


ushachar:
Anything I can do to help with merging this? This issue is forcing us to create a patched version of the plugin whenever an update is released....

artkoshelev:
@elliot-nelson @ushachar i don't think we could merge your changes as is, because it will brake existing behavior. I would like to see this as a setting in plugin configuration, so we could have both workflows.

And talking in general, why not just kill previously running builds when you pushing new changes? You don't care about their result anyway.

elliot-nelson:
@artkoshelev I think that depends on the configuration of the job - one of our jobs, for example, is started by multiple users on their personal branches, so every run is unique and important (it's not true that build #299 is invalidated by build #300, because they are different feature branches requested by different users).

I think for a pipeline workflow, you'd expect those to be auto-generated separate jobs (based on branch), but we use this job to gate access to physical devices.

Another common use case is reporting jobs - we have many jobs that users run to do long running report generation (~20 minutes) - any errors in these jobs should file tickets to appropriate teams, but they are all equally important builds.

ushachar:
@artkoshelev  - Our use case is similar to @elliot-nelson.

Wouldn't the behavior you're describing be better served by configuring the max number of concurrent builds for the job (there's a plugin to do exactly this)?

pieterpaladin:
This PR would solve our concurrency issues as well.

Would vote to merge this.

artkoshelev:
@olamy what do you think about this?

PR #126, opened 2017-08-02
Use modern dependencies

i386:
* Jackson is a lot older than what is used in most Jenkins plugins
* httpcomponents is also a lot loder than what is used elsewhere
** Let maven figure out the right version of `httpcore`
* Branch API 2.0
* Mailer and matrix have had security releases
* Update to latest of misc test libraries used

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/jira-plugin/126)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -35,6 +35,7 @@
     <gmaven.version>1.5</gmaven.version>
     <cobertura.version>2.7</cobertura.version>
     <surefire.version>2.19.1</surefire.version>
+    <jackson.version>2.8.9</jackson.version>
   </properties>
 
   <profiles>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -257,7 +253,7 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>mailer</artifactId>
-      <version>1.15</version>
+      <version>1.11</version>
```
artkoshelev: is this downgrade?

pom.xml
```diff
@@ -267,12 +263,12 @@
     <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
       <artifactId>jackson-core</artifactId>
-      <version>2.6.0-rc3</version>
+      <version>2.8.9</version>
```
artkoshelev: let's replace it with `${jackson.version}` as well

pom.xml
i386: Oops - fixed

pom.xml
i386: Nice catch. Done.

i386:
Thanks for the review @artkoshelev :)

PR #127, opened 2017-08-03
JiraJobAction is not set for job names that are URL encoded

i386:
Pipeline will URL encode job names if the branch or PR name contains a '/' or other non-URL safe characters.

For example a branch will be called "feature/EXAMPLE-123" but the job name would be encoded as "feature%2FEXAMPLE-123". 

In this scenario the search for the JIRA issue key would fail and the action would not be set.

---
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/jira-plugin/127)

### DIFF
--- a/src/main/java/hudson/plugins/jira/JiraJobAction.java
+++ b/src/main/java/hudson/plugins/jira/JiraJobAction.java
@@ -18,6 +18,7 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.io.IOException;
 import java.net.URL;
+import java.net.URLDecoder;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Matcher;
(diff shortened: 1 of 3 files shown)


i386:
@artkoshelev here is a small bug I found in my own contribution from a few months ago.

src/test/java/hudson/plugins/jira/JiraJobActionTest.java
```diff
@@ -15,12 +13,8 @@
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
-import static org.mockito.Matchers.any;
 import static org.mockito.Matchers.anyObject;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.never;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
artkoshelev: please don't use wildcard imports

i386: Sorry about that - new machine and I still don't have IntelliJ setup right again!

src/main/java/hudson/plugins/jira/JiraJobAction.java
```diff
@@ -83,19 +83,24 @@ public static void setAction(@Nonnull Job<?, ?> job, @Nonnull JiraSite site) thr
...
+                issueKey = m.group(1);
+                break;
             }
         }
 
         if (issueKey != null) {
             JiraIssue issue = site.getIssue(issueKey);
-            job.addAction(new JiraJobAction(job, issue));
-            job.save();
+            if (issue != null) {
```
i386: Found a small NPE here

warden: Shouldn't we log a WARN in this case? (the issue pattern appeared, but the issue wasn't found for some reason - hence NPE)

i386: That could happen if someone types `JENKIN-123` instead of `JENKINS-123` which is more likely the case than the JIRA ticket being missing?

warden: There is a number of reasons this could happen, that's why it's good  to see it at least in INFO and it doesn't hurt the logs :)
- when no permissions to issue
- when issue was deleted by the admin
- when somebody made a typo

PR #128, opened 2017-08-04
JENKINS-45992 : Adding missing runtime atlassian fugue dependency

bguerin:
This change is [Reviewable](https://reviewable.io/reviews/jenkinsci/jira-plugin/128)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,6 +20,7 @@
     <mockito.version>1.10.19</mockito.version>
     <powermock.version>1.6.3</powermock.version>
     <jira-rest-client.version>4.0.0</jira-rest-client.version>
+    <fugue.version>2.6.1</fugue.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -223,6 +224,11 @@
       </exclusions>
     </dependency>
     <dependency>
+      <groupId>com.atlassian.fugue</groupId>
+      <artifactId>fugue</artifactId>
+      <version>${fugue.version}</version>
+    </dependency>
```
warden: Shouldn't we use scope: runtime for this dependency?

bguerin: Only considering Maven, yes. But I am not sure if it will be embedeed in the hpi / jpi if we do so ...

bguerin: Tested with runtime, works, updating PR

PR #129, opened 2017-08-16
[JENKINS-46243] Failing build on TimeoutException

kuldazbraslav:
Fix for JENKINS-46243. Fails the build step when TimeoutException occurs instead of silently finishing.

### DIFF
--- a/src/main/java/hudson/plugins/jira/JiraRestService.java
+++ b/src/main/java/hudson/plugins/jira/JiraRestService.java
@@ -45,6 +45,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import java.util.logging.Logger;
 
 import static java.util.logging.Level.WARNING;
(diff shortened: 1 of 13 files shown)


artkoshelev:
Couple of questions here:
Why there are some old commits from 10 Mar?
Can you please add test covering new logic?

kuldazbraslav:
@artkoshelev : 
Hi,
I started from v2.3.1, as that was the version I had primarily troubles with.  Thus I created my forked maintenance branch I use now within my Jenkins instance. Then I merged the changes into master. So that's probably reason for the old commits. If it should have been done in a different way, I can change it.

I can try adding the tests.

warden:
Please rebase the PR against the master branch and force push, generally all hotfixes should be done on the current master branch.

src/main/java/hudson/plugins/jira/JiraRestService.java
```diff
@@ -72,7 +73,7 @@
     private final String authHeader;
 
     private final String baseApiPath;
-
```
warden: I understand that you invoked a default formatter against the whole file - that's OK, but please split that formatting into a separate commit. It's hard to review the whole PR this way + in the history review it's easier to do "git blame" later to see when a line was changed and why.

kuldazbraslav:
Rebased onto master branch; whitespace changes removed.

I'm kind of afraid writing any new test cases, since I added in fact no new functionality.

artkoshelev:
> I'm kind of afraid writing any new test cases, since I added in fact no new functionality.
you've changed current behavior, so it needs to be documented in unit-test

PR #130, opened 2017-09-18
Various POM mistakes

jglick:
@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,7 +33,6 @@
     
     <!-- tests -->
     <groovy.version>2.4.5</groovy.version>
-    <gmaven.version>1.5</gmaven.version>
     <cobertura.version>2.7</cobertura.version>
     <surefire.version>2.20</surefire.version>
     <jackson.version>2.8.9</jackson.version>
@@ -144,7 +143,6 @@
       <plugin>
         <groupId>org.codehaus.gmaven</groupId>
         <artifactId>gmaven-plugin</artifactId>
-        <version>${gmaven.version}</version>
         <configuration>
           <source>1.${java.level}</source>
           <providerSelection>1.8</providerSelection>
@@ -153,7 +151,7 @@
           <dependency>
             <groupId>org.codehaus.gmaven.runtime</groupId>
             <artifactId>gmaven-runtime-1.8</artifactId>
-            <version>${gmaven.version}</version>
+            <version>1.5</version>
           </dependency>
           <dependency>
             <groupId>org.codehaus.groovy</groupId>
@@ -192,16 +190,6 @@
   		<version>2.3.3</version>
 	</dependency>
     <dependency>
-      <groupId>javax.activation</groupId>
-      <artifactId>activation</artifactId>
-      <version>1.1.1</version>
-    </dependency>
-    <dependency>
-      <groupId>commons-discovery</groupId>
-      <artifactId>commons-discovery</artifactId>
-      <version>0.5</version>
-    </dependency>
-    <dependency>
       <groupId>com.atlassian.jira</groupId>
       <artifactId>jira-rest-java-client-api</artifactId>
       <version>${jira-rest-client.version}</version>
@@ -221,6 +209,14 @@
           <groupId>com.google.guava</groupId>
           <artifactId>guava</artifactId>
         </exclusion>
+        <exclusion>
+          <groupId>org.springframework</groupId>
+          <artifactId>spring-beans</artifactId>
+        </exclusion>
+        <exclusion>
+          <groupId>javax.activation</groupId>
+          <artifactId>activation</artifactId>
+        </exclusion>
       </exclusions>
     </dependency>
     <dependency>
@@ -238,6 +234,12 @@
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>httpclient</artifactId>
       <version>4.5.3</version>
+      <exclusions>
+        <exclusion>
+          <groupId>commons-codec</groupId>
+          <artifactId>commons-codec</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.apache.httpcomponents</groupId>
@@ -261,11 +263,6 @@
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>mailer</artifactId>
       <version>1.20</version>
-    </dependency>
-    <dependency>
-      <groupId>commons-codec</groupId>
-      <artifactId>commons-codec</artifactId>
-      <version>1.10</version>
     </dependency>
     <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
@@ -302,6 +299,10 @@
           <exclusion>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-simple</artifactId>
+          </exclusion>
+          <exclusion>
+            <groupId>commons-codec</groupId>
+            <artifactId>commons-codec</artifactId>
           </exclusion>
         </exclusions>
     </dependency>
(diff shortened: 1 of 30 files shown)


pom.xml
```diff
@@ -25,15 +26,13 @@
 
     <!-- jenkins -->
     <jenkins.version>1.642.3</jenkins.version>
-    <hpi-plugin.version>1.95</hpi-plugin.version>
```
jglick: must not override

pom.xml
```diff
@@ -63,7 +62,7 @@
     <connection>scm:git:git://github.com/jenkinsci/jira-plugin.git</connection>
     <developerConnection>scm:git:git@github.com:jenkinsci/jira-plugin.git</developerConnection>
     <url>https://github.com/jenkinsci/jira-plugin</url>
-    <tag>jira-2.4.1</tag>
+    <tag>HEAD</tag>
```
jglick: `maven-release-plugin` snafu?

pom.xml
```diff
@@ -144,7 +143,6 @@
       <plugin>
         <groupId>org.codehaus.gmaven</groupId>
         <artifactId>gmaven-plugin</artifactId>
-        <version>${gmaven.version}</version>
```
jglick: inherit from parent to fix some issues

pom.xml
```diff
@@ -192,16 +190,6 @@
   		<version>2.3.3</version>
 	</dependency>
     <dependency>
-      <groupId>javax.activation</groupId>
-      <artifactId>activation</artifactId>
```
jglick: Version from Jenkins core will be used instead.

pom.xml
```diff
@@ -192,16 +190,6 @@
...
-      <version>1.1.1</version>
-    </dependency>
-    <dependency>
-      <groupId>commons-discovery</groupId>
-      <artifactId>commons-discovery</artifactId>
```
jglick: ditto

pom.xml
```diff
@@ -221,6 +209,14 @@
           <groupId>com.google.guava</groupId>
           <artifactId>guava</artifactId>
         </exclusion>
+        <exclusion>
+          <groupId>org.springframework</groupId>
+          <artifactId>spring-beans</artifactId>
```
jglick: ditto

pom.xml
```diff
@@ -263,11 +265,6 @@
       <version>1.20</version>
     </dependency>
     <dependency>
-      <groupId>commons-codec</groupId>
-      <artifactId>commons-codec</artifactId>
```
jglick: ditto

pom.xml
```diff
@@ -303,13 +300,23 @@
...
     </dependency>
     <dependency>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>p4</artifactId>
         <version>1.3.8</version>
         <optional>true</optional>
+        <exclusions>
+            <exclusion>
+                <groupId>org.apache.commons</groupId>
+                <artifactId>commons-compress</artifactId>
```
jglick: ditto

pom.xml
```diff
@@ -356,6 +363,26 @@
       <classifier>tests</classifier>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.objenesis</groupId>
+      <artifactId>objenesis</artifactId>
```
jglick: dependency upper bound

pom.xml
```diff
@@ -356,6 +363,26 @@
...
+      <version>2.1</version>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>junit</artifactId>
```
jglick: ditto

pom.xml
```diff
@@ -356,6 +363,26 @@
...
+      <version>1.3</version>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>script-security</artifactId>
```
jglick: ditto

pom.xml
```diff
@@ -356,6 +363,26 @@
...
+      <version>1.16</version>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>structs</artifactId>
```
jglick: ditto

src/main/java/hudson/plugins/jira/RunScmChangeExtractor.java
```diff
@@ -25,7 +26,8 @@ private RunScmChangeExtractor() {
 
     public static List<ChangeLogSet<? extends Entry>> getChanges(Run<?, ?> run) {
         if (run instanceof AbstractBuild) {
-            return ((AbstractBuild<?, ?>) run).getChangeSets();
```
jglick: inlined restricted method, pending core update to IIRC 2.60.x when you can use the official API

pom.xml
warden: um, why removing the properties version definition but then actually you define it in dependencies?
It's better to keep all versions in central place I believe.

pom.xml
warden: Kind of, that was just the release process I was following. We can switch to HEAD though, I don't mind.

pom.xml
```diff
@@ -356,6 +363,26 @@
...
+      <version>1.4</version>
```
warden: Version definition in central place (or inherited from parent?)

src/main/java/hudson/plugins/jira/RunScmChangeExtractor.java
warden: Is this related to the POM fixes?

pom.xml
warden: This was purpose, I remember I was trying to bump it but there were some docs/lint checks IIRC that made the build failed (and indeed the build is now failing :-) )

pom.xml
```diff
@@ -4,7 +4,8 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.17</version>
+    <version>2.33</version>
```
recena: 👍

recena: I would take advantage this PR for updating the parameter `<url>`for using `https`

pom.xml
recena: @jglick We could remove `<project.build.sourceEncoding>` since it is inherited as well.

jglick: > indeed the build is now failing

Because you were trying to build using JDK 7, which is no longer supported. (And not necessary.)

pom.xml
jglick: Will do.

pom.xml
jglick: > We could remove `<project.build.sourceEncoding>` since it is inherited as well.

Right, will do.

pom.xml
jglick: So, if you *are* using `maven-release-plugin`, then the value of `tag` in regular branch versions should always be `HEAD`; it is only set to a tag name during the tagged commit, then reset during the commit which switches back to a snapshot.

pom.xml
jglick: Exactly. I am picking up a different version for `gmaven-plugin` from the parent; then keeping the version of `gmaven-runtime-1.8` you had (since the `-jenkins-3` patches only affect the plugin not the runtime). Since there is now only one use of `1.5`, I inlined the property.

pom.xml
jglick: Neither. Selecting a version newer than what you would otherwise get from transitive dependencies. https://github.com/jenkins-infra/jenkins.io/pull/1130 will go over this when completed.

src/main/java/hudson/plugins/jira/RunScmChangeExtractor.java
jglick: Yes. When you update the POM, your plugin (correctly) fails to compile. See https://github.com/jenkinsci/plugin-pom/pull/61.

PR #131, opened 2017-09-20
Switch to a more up-to-date domain name for Atlassian deps

rtyler:
The certificate for m2proxy.atlassian.com has expired today, and the best I can tell is that the domain is deprecated in favor of maven.atlassian.com

Monkey business


rtyler:
@warden I hope you don't mind, I'm going to merge this to unbreak the master builds.

PR #138, opened 2018-01-25
JENKINS-45933 fix warning messages when creating a JiraIssueUpdater

grecoj:
After updating my jenkins and jira plugin versions, when creating an instance of JiraIssueUpdater inside my pipeline code I encountered https://issues.jenkins-ci.org/browse/JENKINS-45933

Example code that produces the issue:
`step([$class: 'hudson.plugins.jira.JiraIssueUpdater',
              issueSelector: [$class: 'hudson.plugins.jira.selector.ExplicitIssueSelector', issueKeys: jiraTicket],
              scm: [$class: 'GitSCM',
                    branches: [[name: issueBranch]],
                    userRemoteConfigs: [[url: 'ssh://git@source.git', credentialsId: 'user']],
                    extensions: []],
              labels: null])`

From what I can tell this issue arises because instead of using the provided constructor the pipeline code instead tries to use the no arg constructor and then use setters to modify the object state. I could be wrong but I think it's the change in groovy from 2.7 to 2.11 which changes how the interaction with private variables works. Adding getter and setter methods to this class has resolved the issue for me.

### DIFF
--- a/src/main/java/hudson/plugins/jira/JiraIssueUpdater.java
+++ b/src/main/java/hudson/plugins/jira/JiraIssueUpdater.java
@@ -74,6 +74,27 @@ public class JiraIssueUpdater extends Recorder implements MatrixAggregatable, Si
         return BuildStepMonitor.BUILD;
     }
 
+    public void setIssueSelector(AbstractIssueSelector issueSelector) {
+        this.issueSelector = issueSelector;
+    }
+
+    public SCM getScm() {
+        return scm;
+    }
+
+    public void setScm(SCM scm) {
+        this.scm = scm;
+    }
+
+    public List<String> getLabels() {
+        return labels;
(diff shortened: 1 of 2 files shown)


warden:
Thanks for contributing Jason. Could you try adding a test for this change?

rompic:
does this also fix https://issues.jenkins-ci.org/browse/JENKINS-48695 ? thanks for the information.

grecoj:
@warden updated with tests. After some testing I also found that the setter methods were not required to prevent the warning so I removed them.

@rompic unfortunately I was unable to reproduce your issue. I have the same jenkins/plugin versions and I tried using the "dev" and "test" labels while updating which worked successfully for me. Your code looks fine to me as well so I'm not sure what's happening there.

PR #139, opened 2018-01-30
 Partial fix for JENKINS-48357, reverting commit 420f69076b8606c74cf24f301b9446778d88658f

gmshake:
Temporary fix for [JENKINS-48357](https://issues.jenkins-ci.org/browse/JENKINS-48357) use same dependencies as jira-2.5, tested with Jenkins 1.642.3
@oleg-nenashev

### DIFF
--- a/src/main/java/hudson/plugins/jira/JiraSite.java
+++ b/src/main/java/hudson/plugins/jira/JiraSite.java
@@ -803,9 +803,6 @@ public class JiraSite extends AbstractDescribableImpl<JiraSite> {
             site.setTimeout(timeout);            
             try {
                 JiraSession session = site.createSession();
-                if (session == null) {
-                    return FormValidation.error("Failed to login to JIRA");
-                }
                 session.getMyPermissions();
                 return FormValidation.ok("Success");
             } catch (RestClientException e) {
--- a/src/main/java/hudson/plugins/jira/JiraSite.java
+++ b/src/main/java/hudson/plugins/jira/JiraSite.java
@@ -803,6 +803,9 @@ public class JiraSite extends AbstractDescribableImpl<JiraSite> {
             site.setTimeout(timeout);            
             try {
                 JiraSession session = site.createSession();
+                if (session == null) {
+                    return FormValidation.error("Failed to login to JIRA");
+                }
                 session.getMyPermissions();
                 return FormValidation.ok("Success");
             } catch (RestClientException e) {
--- a/pom.xml
+++ b/pom.xml
@@ -35,6 +35,7 @@
     <cobertura.version>2.7</cobertura.version>
     <surefire.version>2.20</surefire.version>
     <jackson.version>2.8.10</jackson.version>
+    <httpclient.version>4.5.3</httpclient.version>
   </properties>
 
   <profiles>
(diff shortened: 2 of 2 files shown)


oleg-nenashev:
It is not going to fix the issue for sure. There is a binary compat issue
when you run other plugins.

OTOH it makes the plugin self-consistent at least. So I would rather +1
shipping it as a temporary fix

src/main/java/hudson/plugins/jira/JiraSite.java
```diff
@@ -803,6 +803,9 @@ public FormValidation doValidate(@QueryParameter String userName,
             site.setTimeout(timeout);            
             try {
                 JiraSession session = site.createSession();
+                if (session == null) {
+                    return FormValidation.error("Failed to login to JIRA");
```
warden: Won't that hide the actual reason it failed to login (because of not showing the stacktrace) as lines 812 and 815 should handle those cases and additionally provide a stacktrace to the output logs with the underlying cause?

gmshake: JiraSite.createSession() might return null session when username or password is not provided. Throwing NullpointerException seems not that helpful to end user but developers.
OK, I'd revert this before a better solution.

michaelneale:
@jglick what do you mean by

>So unless a later version of the JIRA client stops abusing Maven

You mean the atlassian JIRA client or the JIRA plugin? Trying to work out what is most broken here, but clearly this is bad bad bad McBad (well not this PR per se, but the JIRA plugin causes an outside amount of pain).

gmshake:
@jglick This issue is not related to duplicate classes, it happens at runtime, as I commented on [https://issues.jenkins-ci.org/browse/JENKINS-48357](https://issues.jenkins-ci.org/browse/JENKINS-48357) 

> The class org.apache.http.impl.client.cache.CachingHttpAsyncClient tries to instantiate package visible class org.apache.http.impl.client.cache.BasicHttpCache. It is OK in normal case, but fails if CachingHttpAsyncClient and BasicHttpCache are loaded from different classloaders, becoming two different run-time packages, causing throwing java.lang.IllegalAccessError.

>We unfortunately encounter this abnormal case, that Jira-plugin depends on httpcomponents-asyncclient loaded by plugin classloader and httpclient-cache loaded by another(which loads apache-httpcomponents-client-4-api).

Maven and Maven plugins do NOT help with such case that Maven treat all dependencies as in the same classpath and are loaded into a single "flat" classloader.

gmshake:
@oleg-nenashev @jglick 
There are newer versions of JRJC, the latest is 5.0.4. Since 5.0.0 atlassian-httpclient depended by JRJC was upgrade from 0.23.0 to 1.0.0 and no long bundles apache httpclient stuff. 
I'd think it was better maintainable than JRJC 4.0.0. Any plans to upgrade JRJC to latest version?

See [https://bitbucket.org/atlassian/jira-rest-java-client/commits/c30a1ec3192972200b2662c890a41683d153ea4a](https://bitbucket.org/atlassian/jira-rest-java-client/commits/c30a1ec3192972200b2662c890a41683d153ea4a)

michaelneale:
@gmshake those newer libs without the bundling of http client may be worth a try?

gmshake:
@michaelneale I tried to compile JIRA plugin with JRJC 5.0.4 but failed to pass EnforceBytecodeVersion enforcer rule. Newer JRJC libs are targeted to JDK 1.8, and do not meet the requirement. :(

michaelneale:
@gmshake you could try changing https://github.com/jenkinsci/jira-plugin/blob/master/pom.xml#L19 to 8 - but may require the jenkins core version to be bumped (this may not be a bad idea anyway - that version is old and dangerous). At least worth trying.

gmshake:
@michaelneale It compiles when bumping maxJdkVersion to 1.8 :)
I can confirm that upgrade JRJC to 5.0.4 solely does not resolve the issue. Same exception as JENKINS-48357  if including dependency apache-httpcomponents-client-4-api. Remove dependency apache-httpcomponents-client-4-api and everything looks good to me.

gmshake:
@oleg-nenashev @jglick @michaelneale 

> Since 5.0.0 atlassian-httpclient depended by JRJC was upgrade from 0.23.0 to 1.0.0 and no long bundles apache httpclient stuff.

To be clear, what i meant was atlassian-httpclient bundles apache http client in 0.23.0 distribution but not in 1.0.0.  atlassian-httpclient 1.0.0 still depends on apache http client.

warden:
IIURC, removing `apache-httpcomponents-client-4-api` will bring us to the original state of `2.5` where the original issue described in JENKINS-48357 still will occasionally happen, but we won't get a massive classloader conflict coming from `apache-httpcomponents-client-4-api` ?

If so, then I vote for merging this one to unblock/cleanup `master`, release it with `2.5.2` (to get other changes from 2.5.1 out) and pick up the work of refactoring `httpcomponents` usage and/or upgrading JRJC client in branch/pr?

gmshake:
@warden Exactly :)

gmshake:
@oleg-nenashev @jglick @michaelneale @warden I'm closing this in favor of https://github.com/jenkinsci/jira-plugin/pull/147 and further steps on the issue https://github.com/jenkinsci/jira-plugin/pull/145 . Thanks for your job :)

michaelneale:
thank you @gmshake - will be good to see the back of this hey!

PR #142, opened 2018-03-08
Fix for JENKINS-44989 Updating JIRA-issues deadlock

ttat:
This was incorrectly refactored into a recursive call.

This was changed from:
https://github.com/jenkinsci/jira-plugin/commit/2eace064c77979cab041e61d82dea47eb44cc886#diff-e381acd1474876247543cc9edc4ea09eL77

Into this:
https://github.com/jenkinsci/jira-plugin/commit/2eace064c77979cab041e61d82dea47eb44cc886#diff-e381acd1474876247543cc9edc4ea09eR134

If the detected changes happen to contain a lot of dependency changes, which can also detect a lot of dependency changes, the build can take a very long time.
JENKINS-46995 makes this bug more noticeable.


artkoshelev:
LGTM

PR #144, opened 2018-03-28
JENKINS-50371 stop limiting number of issues to 50

holgerfriedrich:


artkoshelev:
Could we make it configurable instead?

warden:
@artkoshelev I suggested the same in the JIRA ticket, but this is a bigger "feature" to do I guess. This PR only makes it consistent with other places in the code to not limit the fetched issues numbers on the plugin side.
I'd merge this one and we can open a new feature/improvement request for this not to be forgotten?

PR #145, opened 2018-05-01
[JENKINS-48357] get rid of atlassian http client

olamy:
a first step with importing atlassian httpclient sources (still some pruning to do.... but good start)
At least now we are not anymore dependant on a third party regarding Apache httpclient version used.


artkoshelev:
btw had you tried to update to latest JRJC or enforce atlassian-http-client-plugin version to latest?

vivek:
@olamy LGTM. Did you get to do any testing, I guess if you could reproduce the class loading error then its easier to see if it fixes those issues. I was able to reproduce it in my env with blueocean, I can test it out with this PR. Maybe you can deploy a versioned snapshot that I could try it in my env?

artkoshelev:
code style is out of sync with the rest of the code...

pom.xml
```diff
@@ -217,19 +217,29 @@
...
-      <scope>runtime</scope>
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>apache-httpcomponents-client-4-api</artifactId>
       <version>4.5.3-2.0</version>
     </dependency>
+
+    <dependency>
+      <groupId>org.apache.httpcomponents</groupId>
```
michaelneale: Shoudl this perhaps be https://wiki.jenkins.io/display/JENKINS/Async+Http+Client+Plugin cc @vivek

michaelneale:
can ignore the coverage - as that is just bringing over code that is already in use (surprised it only went down that little). 

I think maybe the async dependency should be the plugin, but otherwise looks ok - would like @vivek to take a look. Is there a way to confirm this helps? (I guess we will know when it is out there). 

very nice!

pom.xml
olamy: This one is based on 
```
      <groupId>com.ning</groupId>
      <artifactId>async-http-client</artifactId>
      <version>1.9.40</version>
```

michaelneale: ah right - maybe I was thinking of https://wiki.jenkins.io/display/JENKINS/Apache+HttpComponents+Client+4.x+API+Plugin - does that include async?

olamy:
@artkoshelev temporary not sure :-) upgrading to new version of JRJC will fix now but this problem can(will?) happen again in few months in case of non sync version again.  
Code coverage can definitely be improved by testing the httpclient (I can add some tests if needed).
@vivek @michaelneale ideally once this one https://github.com/jenkinsci/apache-httpcomponents-client-4-api-plugin/pull/6 is merged some dependencies can be removed here (well might need some blueocean as well)

michaelneale:
ah @olamy you just answered question... right would be good to get https://github.com/jenkinsci/apache-httpcomponents-client-4-api-plugin/pull/6 in, but it looks like async is not in there yet, so it shouldn't be a problem for now for this.

olamy:
@vivek not sure how/where to deploy a snapshot version?

artkoshelev:
It looks like it requires some common approach across plugins, e.g. use only `<provided>` scope for common low-level components like `apache-http-client` and let them be defined in jenkins core. @oleg-nenashev ?

michaelneale:
@olamy a snapshot of this? I believe when @abayer made me do it `mvn deploy` would just do it (ie don't release...). Another option is to set the version to beta and it will go to experimental UC (not sure which open is better, and if they need to be done once it is merged).

pom.xml
gmshake: @michaelneale https://wiki.jenkins.io/display/JENKINS/Apache+HttpComponents+Client+4.x+API+Plugin does not include async part right now.

olamy:
@vivek you can try with https://repo.jenkins-ci.org/snapshots/org/jenkins-ci/plugins/jira/2.5.2-SNAPSHOT/jira-2.5.2-20180501.075458-1.hpi , I tested locally with blueocean installed.
@artkoshelev yup that's the goal but ATM async is not yet part of the bundle.
@michaelneale yup mvn deploy I sometimes forget how maven is so simple :-)

pom.xml
gmshake: @olamy 
1. The 4.1.3 version of org.apache.httpcomponents:httpasyncclient does not have any package level dependency on other part of httpcomponents, but since Apache httpcomponents maintainers have freedom to change the design, can you confirm that this will not happen in future? Or another java.lang.IllegalAccessError might occur when httpasyncclient upgraded.
2. Rewrite HTTP staff to use https://wiki.jenkins.io/display/JENKINS/Async+Http+Client+Plugin would be great :)

dwnusbaum:
@olamy Just a heads up that `apache-httpcomponents-client-4-api-plugin:4.5.5-2.0` has been released with jenkinsci/apache-httpcomponents-client-4-api-plugin#6 which adds `httpasyncclient` and `httpasyncclient-cache` to the plugin.

vivek:
Yay, thanks @dwnusbaum for pr and release of `apache-httpcomponents-client-4-api-plugin:4.5.5-2.0` 👏

olamy:
an other reason to import Atlassian httpclient sources is to be able to fix proxy support (https://issues.jenkins-ci.org/browse/JENKINS-25829).

pom.xml
gmshake: @olamy No problem now since https://github.com/jenkinsci/apache-httpcomponents-client-4-api-plugin/pull/6 has been merged. 👍

olamy:
@vivek fair point regarding coverage ;-)

michaelneale:
@artkoshelev I am not sure there are tests 😬

olamy:
@artkoshelev I know it's not an ideal solution but well we are already blind with what's going on... I'd like to propose some shading of Apache httpcomponents within the http layer of the jira client http layer to avoid future compatibility issues. But It seems to be closed sources....
The pom here https://packages.atlassian.com/maven-external/com/atlassian/httpclient/atlassian-httpclient-parent/0.23.0/atlassian-httpclient-parent-0.23.0.pom mentioned https://bitbucket.org/atlassian/atlassian-httpclient but it's private access...

vivek:
Awesome, thanks @olamy 🙌

michaelneale:
nice one @olamy  - this bug is almost as painful as running 100km in 1 day.

EliaBracciSumo:
Hi @olamy! I need to use [this](https://github.com/apache/httpcomponents-client/blob/48e4229843eed904858206d650da8f5090f40d0b/httpclient5/src/main/java/org/apache/hc/client5/http/entity/mime/MultipartEntityBuilder.java) class to call `addAttachment` function. Since apache httpclient has been removed, should I import all needed files inside `com.atlassian.httpclient.apache.httpcomponents` package inside the project? Thank you in advance!

PR #150, opened 2018-05-16
[JENKINS-51312] upgrade core to 2.60.1

olamy:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
   </parent>
 
   <artifactId>jira</artifactId>
-  <version>2.5.3-SNAPSHOT</version>
+  <version>3.0.0-SNAPSHOT</version>
   <packaging>hpi</packaging>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -388,7 +388,13 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>structs</artifactId>
-      <version>1.4</version>
+      <version>1.7</version>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>credentials</artifactId>
```
artkoshelev: why do we need this now?

olamy: it is a dependency to force because of upgrade.
Please note the test scope.

artkoshelev: 👍

PR #159, opened 2018-07-08
Move progressMatchingIssues out of JiraSite

artkoshelev:


warden:
I like this approach for the decomposing and cleanup part :) 
@artkoshelev are you still up for it?
@olamy for the backward compatibility, how about scheduling it for 4.x and do a @Deprecated in a separate PR for the affected method?

artkoshelev:
It would be nice to do a major release removing all code marked `@Deprecated`

olamy:
@artkoshelev why? maybe it's sad but you can break few people so we should avoid that....

artkoshelev:
it should be fine to break backward-cap on major releases, you can't keep legacy code forever, it's becoming more and more harder to maintain

warden:
I'm a believer of the approach that if we want the OSS projects to be alive,non-buggy and up2date with recent (here recent meaning <3y) standards, then we should regularly clean up the code.
That's why [`@Deprecated`](https://docs.oracle.com/javase/7/docs/technotes/guides/javadoc/deprecation/deprecation.html) was created, it's up to us though to define the release cycles policy and when to clean it up I guess.

warden:
@artkoshelev what do you think about rebasing this to `feature/4.0` ?

PR #160, opened 2018-08-09
[JENKINS-52906] Configuration-as-Code compatibility

ndeloof:
Use a `PersistedList` to manage `sites` attributes, which is by design safe to be exposed as public attribute. This allows Configuration-as-Code to manage configuring sites without need to break binary compatibility on `getSites` and `setSites`.

### DIFF
--- a/src/main/java/hudson/plugins/jira/JiraProjectProperty.java
+++ b/src/main/java/hudson/plugins/jira/JiraProjectProperty.java
@@ -5,8 +5,11 @@ import hudson.model.Job;
 import hudson.model.JobProperty;
 import hudson.model.JobPropertyDescriptor;
 import hudson.util.CopyOnWriteList;
+import hudson.util.PersistedList;
 import net.sf.json.JSONObject;
+import org.jenkinsci.Symbol;
 import org.kohsuke.stapler.DataBoundConstructor;
+import org.kohsuke.stapler.DataBoundSetter;
 import org.kohsuke.stapler.Stapler;
 import org.kohsuke.stapler.StaplerRequest;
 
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/plugins/jira/JiraProjectProperty.java
```diff
@@ -85,6 +90,7 @@ public String getDisplayName() {
             return Messages.JiraProjectProperty_DisplayName();
         }
 
+        @Deprecated
```
artkoshelev: please put comment on why it's deprecated and what could be used instead

src/main/java/hudson/plugins/jira/JiraProjectProperty.java
```diff
@@ -66,8 +69,10 @@ public DescriptorImpl getDescriptor() {
     @Extension
     public static final DescriptorImpl DESCRIPTOR = new DescriptorImpl();
 
+    @Symbol("jira")
```
artkoshelev: what is this for?

ndeloof: see https://github.com/jenkinsci/configuration-as-code-plugin/pull/432/files#diff-98519df47b68dd6c5fa65a3101a8d7deL2

symbol is used to give class a "human friendly name". In JCasC context this allow to use an reasonable name for Jira configuration section

ndeloof:
@artkoshelev requested change applied, can you please reconsider review status ?
CI failure is due to previously existing findbugs issues

ndeloof:
@artkoshelev yes indeed
but I'm not confident with the test framework jira-plugin use and can't get how the proposed changes impact this test case.

artkoshelev:
you need to find out what's wrong and fix tests in order to merge changes

ndeloof:
then I will let you handle this compatibility issue as I can't fix this by myslef

casz:
@ndeloof the test is calling JiraProjectProperty.DESCRIPTOR.setSites(site2) which I guess is incorrect and only site 2 is generating errors.

PR #166, opened 2018-09-28
fix some strange http configuration confusion between connect timeout and read timeout

olamy:


parapunzi:
sorry to use this comments for bugreporting: had a strange issue today updating to the released JIRA plugin 3.0.3: in the Jenkins config GUI the new (?) fields 'Read timeout' and 'Thread Executor Size' both had value 0 -- and thus builds failed with 'Failed to parse changelog' and change lists had missing Jira-Links. Setting the fields to 30 and 10 fixed the problem.

PR #172, opened 2018-10-27
use numbers on jira site jelly

casz:


warden:
This seems undocumented on https://wiki.jenkins.io/display/JENKINS/Jelly+form+controls and as well on https://jenkins.io/doc/developer/forms/ ;-) 
What's the advantage of using `f:number` ? I'm guessing that any non-digit will be converted to a numeric(default?).

We still have the default actually defined in jelly and in `public static final int DEFAULT_READ_TIMEOUT = 30;`, do you think one of those can re-use the other?

casz:
advantage of number is I can only type numbers... No text is allowed.

https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/lib/form/number.jelly
it is basically a input type of number. Functions exactly like this: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#Value with default value even.

PR #173, opened 2018-10-27
Simplify constructor, main URL is the only mandatory field

casz:


casz:
Damn travis is slow without those optimistion in #171

casz:
Knew this was gonna happen, now I am a slave to your test coverage 😅

src/main/java/hudson/plugins/jira/JiraSite.java
```diff
@@ -120,12 +120,12 @@
      * URL of JIRA for normal access, like <tt>http://jira.codehaus.org/</tt>.
      * Mandatory. Normalized to end with '/'
      */
-    public final URL alternativeUrl;
+    public URL alternativeUrl;
```
artkoshelev: this still should be final, right?

src/test/java/hudson/plugins/jira/JiraSite2Test.java
```diff
@@ -170,4 +179,137 @@ public void testDeserializeWithoutCredentials() {
...
+        thrown.expect(AssertionError.class);
+        thrown.expectMessage("java.net.MalformedURLException: no protocol");
+        new JiraSite("malformed.url");
+    }
+
+    @Test
+    @WithoutJenkins
+    public void malformedAlternativeUrl() {
+        JiraSite jiraSite = new JiraSite(nonExistentUrl.toExternalForm());
+        thrown.expect(AssertionError.class);
```
artkoshelev: imo:
1. using `@Test(expected = AssertionError.class)` looks clearer, and you don't need exception rule
2. checking exception message is a bit of overkill - you couple your tests with implementation too tight

src/test/java/hudson/plugins/jira/JiraSite2Test.java
```diff
@@ -170,4 +179,137 @@ public void testDeserializeWithoutCredentials() {
...
+        assertNull(jiraSite.getAlternativeUrl());
+        assertNull(url1);
+        assertNull(url2);
+    }
+
+    @Test
+    @WithoutJenkins
+    public void ensureMainUrlIsMandatory() {
+        thrown.expect(AssertionError.class);
+        thrown.expectMessage("URL cannot be null");
```
artkoshelev: it's confusing: why do we say it's null while it's obviously not - it's an empty string?

src/test/java/hudson/plugins/jira/JiraSite2Test.java
```diff
@@ -170,4 +179,137 @@ public void testDeserializeWithoutCredentials() {
...
+        assertNotNull(jiraSite.getCredentialsId());
+        assertNotNull(jiraSite.credentials);
+        assertEquals(cred, jiraSite.getCredentialsId());
+        assertEquals(credentials.getUsername(), jiraSite.credentials.getUsername());
+        assertEquals(credentials.getPassword(), jiraSite.credentials.getPassword());
+    }
+
+    @Test
+    @WithoutJenkins
+    public void gettersAndSetters() throws Exception {
```
artkoshelev: Please don't add dummy test code for the dummy methods in sake of coverage numbers. There is no point in testing getters and setters if they are not doing anything but assigning fields. In theory, they should be covered by tests for related features.

src/test/java/hudson/plugins/jira/JiraSite2Test.java
```diff
@@ -170,4 +179,137 @@ public void testDeserializeWithoutCredentials() {
...
+        thrown.expectMessage("java.net.MalformedURLException: no protocol");
+        jiraSite.setAlternativeUrl("malformed.url");
+    }
+
+    @Test
+    public void credentials() throws Exception {
+        JiraSite jiraSite = new JiraSite(exampleOrg.toExternalForm());
+        jiraSite.setCredentialsId("");
+        assertNull(jiraSite.getCredentialsId());
+        assertNull(jiraSite.credentials);
```
artkoshelev: first 4 lines belongs to a separate test, e.g. `credentialsAreNullByDefault`

src/test/java/hudson/plugins/jira/JiraSite2Test.java
```diff
@@ -170,4 +179,137 @@ public void testDeserializeWithoutCredentials() {
...
+        String cred = "cred-1-id";
```
artkoshelev: `credId`

src/test/java/hudson/plugins/jira/JiraSite2Test.java
```diff
@@ -170,4 +179,137 @@ public void testDeserializeWithoutCredentials() {
...
+        UsernamePasswordCredentialsImpl credentials = new UsernamePasswordCredentialsImpl(
+            CredentialsScope.GLOBAL, cred, null, user, pwd);
+
+        SystemCredentialsProvider systemProvider = SystemCredentialsProvider.getInstance();
+        systemProvider.getCredentials().add(credentials);
+        systemProvider.save();
+        jiraSite.setCredentialsId(cred);
+        assertNotNull(jiraSite.getCredentialsId());
+        assertNotNull(jiraSite.credentials);
+        assertEquals(cred, jiraSite.getCredentialsId());
```
artkoshelev: we don't really care which ID that credentials has. so i would suggest to check username/password against raw `user` and `pwd` values.

src/test/java/hudson/plugins/jira/JiraSite2Test.java
```diff
@@ -170,4 +179,137 @@ public void testDeserializeWithoutCredentials() {
...
+        assertTrue(url1.toExternalForm().endsWith("/"));
+        assertTrue(url2.toExternalForm().endsWith("/"));
+    }
+
+    @Test
+    @WithoutJenkins
+    public void urlNulls() {
+        JiraSite jiraSite = new JiraSite(nonExistentUrl.toExternalForm());
+        jiraSite.setAlternativeUrl(" ");
+        URL url1 = JiraSite.toURL("");
```
artkoshelev: too many checks mixed in one test, please move tests related to `JiraSite.toUrl()` method to separate test cases.

src/test/java/hudson/plugins/jira/JiraSite2Test.java
```diff
@@ -170,4 +179,137 @@ public void testDeserializeWithoutCredentials() {
...
+        URL url1 = JiraSite.toURL(nonExistentUrl.toExternalForm());
+        URL url2 = JiraSite.toURL(exampleOrg.toExternalForm());
+        assertTrue(url1.toExternalForm().endsWith("/"));
+        assertTrue(url2.toExternalForm().endsWith("/"));
+    }
+
+    @Test
+    @WithoutJenkins
+    public void urlNulls() {
+        JiraSite jiraSite = new JiraSite(nonExistentUrl.toExternalForm());
```
artkoshelev: that `nonExistentUrl` variable is rather confusing, should be `validPrimaryUrl` maybe? same with `exampleOrg`

src/test/java/hudson/plugins/jira/JiraSite2Test.java
casz: I'll change the message to say `URL cannot be emtpy`.
But that's why we use `Util.fixEmptyAndTrim` so we get nulls instead of dealing with empty string values.

src/main/java/hudson/plugins/jira/JiraSite.java
casz: In Jenkins terms, final is used for mandatory values.
if it were Mandatory it should be final and be in the constructor.

see:
https://jenkins.io/doc/developer/plugin-development/pipeline-integration/#constructor-vs-setters

src/test/java/hudson/plugins/jira/JiraSite2Test.java
casz: Well I added a lot of getters and setters. The only way I could get this pass the coveralls percentage checker.

timja:
@casz you still interested in fixing this up? would be great to get this plugin sorted

casz:
Your welcome to send a PR @timja 😅 I doubt this PR will get past the nazi coveralls checker

timja:
Outstanding changes applied here:
https://github.com/jenkinsci/jira-plugin/pull/177

in 
https://github.com/jenkinsci/jira-plugin/pull/177/commits/d5f8f91bc890189ae3d972dbdc3c4025ce693eaf

casz:
superseded by #177

PR #175, opened 2019-02-12
[JENKINS-55787] Switch labels from entry to checkbox

jsoref:
[https://issues.jenkins-ci.org/browse/JENKINS-55787](https://issues.jenkins-ci.org/browse/JENKINS-55787)

Basically checkboxes should have labels, splitting the labels away from their checkboxes is poor UX, poor accessibility. This change brings the layout more inline with how normal settings UIs lay out checkboxes.


olamy:
thanks guys but please do not miss to update the jenkins issue with the version label :)

PR #176, opened 2019-02-28
open a new tab for jira , from the jenkins changelog page

waterings:
open a new tab for jira , from the jenkins changelog page

### DIFF
--- a/src/main/java/hudson/plugins/jira/JiraChangeLogAnnotator.java
+++ b/src/main/java/hudson/plugins/jira/JiraChangeLogAnnotator.java
@@ -117,7 +117,7 @@ public class JiraChangeLogAnnotator extends ChangeLogAnnotator {
                     text.addMarkup(m.start(1), m.end(1), "<a href='" + url + "'>", "</a>");
                 } else {
                     text.addMarkup(m.start(1), m.end(1),
-                            String.format("<a href='%s' tooltip='%s'>", url, Util.escape(issue.getSummary())), "</a>");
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/plugins/jira/JiraChangeLogAnnotator.java
```diff
@@ -117,7 +117,7 @@ public void annotate(Run<?, ?> build, Entry change, MarkupText text) {
                     text.addMarkup(m.start(1), m.end(1), "<a href='" + url + "'>", "</a>");
```
warden: :+1: but let's be consistent and open a new window here as well

src/main/java/hudson/plugins/jira/JiraChangeLogAnnotator.java
```diff
@@ -117,7 +117,7 @@ public void annotate(Run<?, ?> build, Entry change, MarkupText text) {
...
                 } else {
                     text.addMarkup(m.start(1), m.end(1),
-                            String.format("<a href='%s' tooltip='%s'>", url, Util.escape(issue.getSummary())), "</a>");
+                            String.format("<a href='%s' tooltip='%s' target='_blank'>", url, Util.escape(issue.getSummary())), "</a>");
```
warden: Can you fix the tests for this change?

warden:
After giving it some more thought, there are no places where we are opening up a new tab, so for consistency I'd keep the current behaviour as in other places.
If a user wants to open new tab for some link, one can click ctrl+link to do so.

PR #178, opened 2019-03-20
[WIP] [JENKINS-50643] http client reuse

olamy:


src/main/java/com/atlassian/httpclient/apache/httpcomponents/DefaultHttpClientFactory.java
```diff
@@ -65,9 +66,18 @@ public void dispose(@Nonnull final HttpClient httpClient) throws Exception
     private HttpClient doCreate(HttpClientOptions options, ThreadLocalContextManager threadLocalContextManager)
     {
         checkNotNull(options);
-        final ApacheAsyncHttpClient httpClient = new ApacheAsyncHttpClient(eventPublisher, applicationProperties, threadLocalContextManager, options);
-        httpClients.add(httpClient);
-        return httpClient;
+        // we create only one http client instance as we don't need more
```
artkoshelev: what would happen if options doesn't match?

src/main/java/hudson/plugins/jira/JiraSite.java
```diff
@@ -450,35 +451,55 @@ protected JiraSession createSession() {
...
-            int nThreads = threadExecutorNumber;
-            if(nThreads<1){
-                LOGGER.warning( "nThreads " + nThreads + " cannot be lower than 1 so use default " + DEFAULT_THREAD_EXECUTOR_NUMBER );
-                nThreads = DEFAULT_THREAD_EXECUTOR_NUMBER;
-            }
-            executorService =  Executors.newFixedThreadPool(
-                nThreads, //
-                new ThreadFactory()
+    private ExecutorService getExecutorService() {
+        if (executorService==null)
```
artkoshelev: if only some dependency-injection framework exist in java... :trollface:

src/main/java/com/atlassian/httpclient/apache/httpcomponents/DefaultHttpClientFactory.java
```diff
@@ -20,7 +20,8 @@
     private final EventPublisher eventPublisher;
     private final ApplicationProperties applicationProperties;
     private final ThreadLocalContextManager threadLocalContextManager;
-    private final Set<ApacheAsyncHttpClient> httpClients = new CopyOnWriteArraySet<ApacheAsyncHttpClient>();
+    // shared http client
+    private static final Set<ApacheAsyncHttpClient> httpClients = new CopyOnWriteArraySet<>();
```
artkoshelev: we don't need a set now if we create only one instance

src/main/java/com/atlassian/httpclient/apache/httpcomponents/DefaultHttpClientFactory.java
olamy: I know but this `su...` since version 2.x (not sure which one :) ) and the introduction of the async rest jira api...
It's was a mistake to not have this common http client since this major change!
Options are really common  except the timeout which is handle with the future timeout so I do not see a major problem...

src/main/java/hudson/plugins/jira/JiraSite.java
olamy: oh yes not sure how to use it here.... But I will have a look

src/main/java/com/atlassian/httpclient/apache/httpcomponents/DefaultHttpClientFactory.java
olamy: correct!

olamy: fixed

olamy:
@artkoshelev why do you merge it???? it was still WIP!!! if it was ready to merge I would change the title and remove the `WIP` label. And btw I would have rebase to have a single commit which would have been easier to revert if needed....
That's definitely not a serious way of working!!!

artkoshelev:
@olamy do you want me to revert it, or would you prefer to work on top of merged changes?

artkoshelev:
also there is a merge commit, so no drama if revert needed

olamy:
leave it as is..... but please take care in the future... I usually merge my prs to master when it's ready...

casz:
FYI https://github.blog/2019-02-14-introducing-draft-pull-requests/ :)

PR #183, opened 2019-04-02
Use display names in integrated build comments

schwjm:
## What is changed:

When posting about an integrated build, use `getFullDisplayName()` instead of relying on default `toString()` behavior which does not consider display names.

## Why am I requesting this change?

The internal Jenkins build numbers don't mean much within the organization, so the Jira comments showing the commonly understood version number we use (e.g. 1.0.0.x) would be more desired. From https://github.com/jenkinsci/jenkins/blob/811583eb6007ffa22f81c312d983816e5f36a670/core/src/main/java/hudson/model/Run.java#L780-L796 it seems to be clear that the behavior is unchanged if custom display names are not set.

### DIFF
--- a/src/main/java/hudson/plugins/jira/Updater.java
+++ b/src/main/java/hudson/plugins/jira/Updater.java
@@ -226,7 +226,7 @@ class Updater {
                             "Integrated in [%2$s|%3$s]\n%4$s" :
                             "Integrated in Jenkins build %2$s (See [%3$s])\n%4$s",
                     jenkinsRootUrl,
-                    build,
+                    build.getFullDisplayName(),
                     Util.encode(jenkinsRootUrl + build.getUrl()),
                     getScmComments(wikiStyle, build, recordScmChanges, jiraIssue));
         else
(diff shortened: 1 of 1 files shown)


artkoshelev:
Could you pleas add tests to cover both cases:

- commonly understood version number
- default fallback behavior

schwjm:
@artkoshelev Could you clarify a bit: perhaps confirmation that the result from `getFullDisplayName()` is included is acceptable? I don't know if this code really cares or should care the specifics of how that method works. The desired behavior I'm asking for is to see something closer to what is displayed to the user in general for builds and jobs and that method seems to fit the bill directly.

It seems like I'd either extend or make a sibling to https://github.com/jenkinsci/jira-plugin/blob/d6e924d05297b1ab14acc65d78ea150a46794f6e/src/test/java/hudson/plugins/jira/UpdaterTest.java#L154-L215 ?

Thanks for your patience as I'm a bit out of my element here.

artkoshelev:
nwm, LGTM
@olamy are you OK to merge this?

PR #188, opened 2019-04-30
WIP [JENKINS-57012] Allow folder scope credential for Jira Site

mat1e:
This branch allow the folder's scope credential when you configure a Jira site on a folder.
Like the site, the credentials created will be visible only on items in the scope of the folder (subitems and subfolder...)

jira_plugin_folder_credential

### DIFF
--- a/src/main/java/hudson/plugins/jira/JiraSite.java
+++ b/src/main/java/hudson/plugins/jira/JiraSite.java
@@ -521,8 +521,9 @@ public class JiraSite extends AbstractDescribableImpl<JiraSite> {
         }
         return jiraSession;
     }
-    
-    private JiraSession getSession() {
+
+    @Deprecated
+    public JiraSession getSession() {
         if (jiraSession == null) {
             StandardUsernamePasswordCredentials credentials = CredentialsHelper.lookupSystemCredentials(credentialsId, alternativeUrl);
             jiraSession = createSession(credentials);
--- a/src/main/java/hudson/plugins/jira/JiraSite.java
+++ b/src/main/java/hudson/plugins/jira/JiraSite.java
@@ -17,7 +17,6 @@ import com.atlassian.sal.api.executor.ThreadLocalContextManager;
 import com.cloudbees.plugins.credentials.CredentialsProvider;
 import com.cloudbees.plugins.credentials.common.StandardUsernameListBoxModel;
 import com.cloudbees.plugins.credentials.common.StandardUsernamePasswordCredentials;
-import com.cloudbees.plugins.credentials.common.UsernamePasswordCredentials;
 import com.cloudbees.plugins.credentials.domains.URIRequirementBuilder;
 import com.google.common.base.Objects;
 import com.google.common.base.Optional;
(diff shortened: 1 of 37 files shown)


warden:
I only took a brief look as the PR touches a lot of files and public methods (binary compatibility?), but I skimmed through the tests and didn't catch a place where this is actually tested. Did I miss a test that verifies that actually the credentials are invisible from another folder that shouldn't see them?

mat1e:
@warden , you're right, I just updated the existings tests but not create ones for this case. 
I tried to test it using the same way than src/test/java/hudson/plugins/jira/DescriptorImplTest.java but I didn't find how to create a Domain with a folder. I keep looking for a way to do this test.

mat1e:
I added the test

mat1e:
> 
> 
> I only took a brief look as the PR touches a lot of files and public methods (binary compatibility?), but I skimmed through the tests and didn't catch a place where this is actually tested. Did I miss a test that verifies that actually the credentials are invisible from another folder that shouldn't see them?

I added the test

src/main/java/hudson/plugins/jira/JiraSite.java
```diff
@@ -510,9 +500,32 @@ public String getName() {
      * @return null if remote access is not supported.
      */
     @Nullable
-    public JiraSession getSession() {
+    public JiraSession getSession(ItemGroup context) {
```
warden: I think we need to keep binary compatibility here, i.e. mark the public constructor as Deprecated and introduce the new one with context.

src/main/java/hudson/plugins/jira/JiraSite.java
```diff
@@ -896,23 +909,15 @@ public JiraIssue getIssue(final String id) throws IOException {
         }
     }
 
-    @Deprecated
```
warden: Leave the deprecated things for a major release (where they will be cleaned up).

src/main/java/hudson/plugins/jira/JiraSite.java
```diff
@@ -921,17 +926,18 @@ public boolean existsIssue(String id) {
...
     }
 
-    /**
+	/**
      * Generates release notes for a given version.
      *
      * @param projectKey the project key
      * @param versionName the version
      * @param filter      Additional JQL Filter. Example: status in (Resolved,Closed)
+	 * @param run
```
warden: indent

src/main/java/hudson/plugins/jira/JiraSite.java
```diff
@@ -980,10 +986,11 @@ public String getReleaseNotesForFixVersion(String projectKey, String versionName
      * @param projectKey The project key
      * @param toVersion  The new fixVersion
      * @param query      A JQL Query
+     * @param build
```
warden: Would be nice to have a short description for this param so that it doesn't stand out from the others ;-)

src/main/java/hudson/plugins/jira/JiraSite.java
mat1e: I don't find the way to get the context in a DataBoundConstructor... Do you have any idea?

casz: if you have a `doFill` method you can use `@AncestorInPath ItemGroup context`
https://github.com/jenkinsci/hashicorp-vault-plugin/blob/8e70b761172b9b0cf72719aa9191d71d5fc01fbf/src/main/java/com/datapipe/jenkins/vault/configuration/VaultConfiguration.java#L150

mat1e: > 
> 
> if you have a `doFill` method you can use `@AncestorInPath ItemGroup context`
> https://github.com/jenkinsci/hashicorp-vault-plugin/blob/8e70b761172b9b0cf72719aa9191d71d5fc01fbf/src/main/java/com/datapipe/jenkins/vault/configuration/VaultConfiguration.java#L150

I need it in constructor if possible.

casz: What he means you need to keep `getSession()` 

I think for simplicity  `getSession()` should call `getSession(ItemGroup context)` with null.

so 
```java
public JiraSession getSession() {
    return getSession(null);
}
```

PR #189, opened 2019-05-03
[JENKINS-32087] Configure JIRA credentials at job level

mat1e:
Allow to set credential at job level like described on https://issues.jenkins-ci.org/browse/JENKINS-32087

set_credential_on_job


src/main/java/hudson/plugins/jira/JiraCreateIssueNotifier.java
```diff
@@ -451,30 +451,40 @@ public FormValidation doCheckProjectKey(@QueryParameter String value) throws IOE
...
+                        for (Priority priority : session.getPriorities()) {
+                            items.add("[" + site.getName() + "] " + priority.getName(), String.valueOf(priority.getId()));
+                        }
                     }
                 }
             }
             return items;
         }
 
-        public ListBoxModel doFillTypeIdItems() {
```
warden: Can you add a short description for this method as it became quite long now? One-liner will be enough :)

src/main/java/hudson/plugins/jira/JiraCreateIssueNotifier.java
```diff
@@ -451,30 +451,40 @@ public FormValidation doCheckProjectKey(@QueryParameter String value) throws IOE
...
             ListBoxModel items = new ListBoxModel().add(""); // optional field
-            for (JiraSite site : JiraGlobalConfiguration.get().getSites()) {
+            if(item instanceof Project) {
+                project = (Project) item;
+                final JiraProjectProperty jiraProjectProperty = (JiraProjectProperty) project.getProperty(JiraProjectProperty.class);
+                JiraSite site = jiraProjectProperty.getSite();
                 JiraSession session = site.getSession();
                 if (session != null) {
                     for (IssueType type : session.getIssueTypes()) {
                         items.add("[" + site.getName() + "] " + type.getName(), String.valueOf(type.getId()));
```
warden: Do you think sorting the list alphabetically makes sense? (or putting the globally defined sites at the top?)

src/main/java/hudson/plugins/jira/JiraCreateIssueNotifier.java
```diff
@@ -274,13 +278,9 @@ JiraSite getSiteForProject(AbstractProject<?, ?> project) {
      */
     private JiraSession getJiraSession(AbstractBuild<?, ?> build) throws IOException {
 
-        JiraSite site = getSiteForProject(build.getProject());
-
-        if (site == null) {
-            throw new IllegalStateException("JIRA site needs to be configured in the project " + build.getFullDisplayName());
-        }
+        final JiraProjectProperty jiraProjectProperty = build.getParent().getProperty(JiraProjectProperty.class);
```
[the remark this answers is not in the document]
mat1e: I don't think it can throw NPE. It's a method from Run class annotate with @Nonnull

src/main/java/hudson/plugins/jira/JiraCreateIssueNotifier.java
mat1e: Maybe I misunderstood this part, but for me the list of issues types and priorities are got from the Jira API. It must use the connexion to Jira configured on the Project properties and not in Global Configuration.

mat1e:
> 
> 
> Some small comments.
> Also note that coverage has dropped as a result of this change, can you take a look why?

I already checked the coverage drop. It is mostly caused by the doTestConnection method on JiraProjectProperty. It is difficult to test because it's from the descriptor and it is static.

warden:
@mat1e since now it is possible to configure credentials on folder level, I think that would work better for large organisations to separate jobs per folders instead of setting sites specifically per job. Can you comment if that solves your usecase as well?

mat1e:
> 
> 
> @mat1e since now it is possible to configure credentials on folder level, I think that would work better for large organisations to separate jobs per folders instead of setting sites specifically per job. Can you comment if that solves your usecase as well?

@olamy, last time I tested credentials on folder with jira-plugin, it didn't works. If it is working now, yes it is better. 
But It wasn't the requestor of [JENKINS-32087](https://issues.jenkins-ci.org/browse/JENKINS-32087).
Maybe you should ask them too.

warden:
@mat1e you are right, it still doesn't work, thanks for pointing it out, I've taken a shot to fix it in #219.

warden:
@olamy do you agree we can close this in favor of the folder-level configuration method and promote it as the better way to manage creds?

olamy:
@warden oh yes I agree

PR #191, opened 2019-05-21
pom.xml updates

warden:


casz:
you inverted the ordered in the success block so no test coverage report is generated when coveralls is run 😓

PR #194, opened 2019-08-20
JENKINS-59001 - IssueSelectorStep should handle NullPointerException caused by having no configured sites

Evildethow:
https://issues.jenkins-ci.org/browse/JENKINS-59001

@olamy are you able to take a look? 

Thanks.


artkoshelev:
How is it even possible? Fresh install?

Evildethow:
@artkoshelev, in this case, it was triggered by https://issues.jenkins-ci.org/browse/JENKINS-57899

artkoshelev:
I don't think we should fix it here, i'm pretty sure it'll fail somewhere else.

olamy:
even if fixing the root cause is more important. it's still good to be NPE free

PR #195, opened 2019-08-20
[JENKINS-57899] - fix saving global config

casz:
fixes JENKINS-57899

### DIFF
--- a/src/main/java/hudson/plugins/jira/JiraGlobalConfiguration.java
+++ b/src/main/java/hudson/plugins/jira/JiraGlobalConfiguration.java
@@ -29,5 +29,6 @@ public class JiraGlobalConfiguration extends GlobalConfiguration {
     @DataBoundSetter
     public void setSites(List<JiraSite> sites) {
         this.sites = sites;
+        save();
     }
 }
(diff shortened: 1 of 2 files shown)


src/main/java/hudson/plugins/jira/JiraGlobalConfiguration.java
```diff
@@ -29,5 +29,6 @@ public JiraGlobalConfiguration() {
     @DataBoundSetter
     public void setSites(List<JiraSite> sites) {
         this.sites = sites;
+        save();
```
casz: https://github.com/jenkinsci/jenkins/blob/01d7f443c9e3178da7222e1dff3fc8cf82bc6943/core/src/main/java/jenkins/model/GlobalConfiguration.java#L35
As noted by the java doc, you have to call save on the data-bound setters.

olamy:
I'd really like to have unit test for that. Do you have time for that?

olamy:
@casz if you can't no worries I will do it

casz:
I'll add them shortly

casz:
@olamy test added :)

olamy:
@casz awesome! thanks!

olamy:
@casz thanks a lot for the quick fix (and the test ;) )

PR #197, opened 2019-08-21
fix travis cache of maven repo

casz:
Hopefully, this fixes the Travis cache. Seems that builds are always downloading 😓


casz:
Yup seems to have fixed the download: https://travis-ci.org/jenkinsci/jira-plugin/builds/574651593

casz:
Cut off 3 minutes and 32 seconds just by adding proper caching 👍

PR #205, opened 2020-01-16
Upgrade to use jcasc test harness

timja:
JCasc had to create a new dependency that wasn't a `tests` classifier because maven doesn't allow you to bring dependencies through classifiers.

Upgrading here so that PCT works again,

Tracking:
https://github.com/jenkinsci/bom/pull/164

Note: We'll need a release for PCT please


warden:
@timja did you take a look at the failing test maybe?

warden:
@timja could you rebase and check with the current revision please?

PR #213, opened 2020-02-23
JENKINS-53794 JENKINS-56987 Bump to jira rest client 5.2.1

rodrigc:
JENKINS-53794 JENKINS-56987 Bump to jira rest client 5.2.0

This is taken from: https://github.com/jenkinsci/jira-plugin/pull/212

See also:
*  https://bitbucket.org/atlassian/jira-rest-java-client/pull-requests/93 
*  https://bitbucket.org/atlassian/jira-rest-java-client/pull-requests/93/jrjc-240-added-support-for-accountid/d

PR #216, opened 2020-03-03
Accept all status that are in the "done" category as being done

t-8ch:
Instead of hardcoding three statuses to be recognized as "done" this PR switches to use the predefined status category "done".

The first commit bumps the dependencies because the status category was not exposed before.
The second commit should also work alone after #213 is merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ See [GitHub releases](https://github.com/jenkinsci/jira-plugin/releases)
 ### Unreleased
 Release date:  _March 2, 2020_
 * Changed all references of JIRA to Jira per [Atlassian branding updates](https://community.atlassian.com/t5/Feedback-Forum-articles/A-new-look-for-Atlassian/ba-p/638077)
+* JiraCreateIssueNotifier: regard all statuses in "done" category as finished
 
 ### 3.0.11 
 Release date:  _Nov 21, 2019_
--- a/src/main/java/hudson/plugins/jira/JiraCreateIssueNotifier.java
+++ b/src/main/java/hudson/plugins/jira/JiraCreateIssueNotifier.java
@@ -1,5 +1,6 @@
 package hudson.plugins.jira;
 
+import com.atlassian.jira.rest.client.api.StatusCategory;
 import com.atlassian.jira.rest.client.api.domain.Issue;
 import com.atlassian.jira.rest.client.api.domain.IssueType;
 import com.atlassian.jira.rest.client.api.domain.Priority;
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -189,22 +189,12 @@
       <groupId>com.atlassian.jira</groupId>
       <artifactId>jira-rest-java-client-api</artifactId>
       <version>${jira-rest-client.version}</version>
-      <exclusions>
```
warden: Are you sure those exclusions can be safely removed? (see comments in pom changes of #213 )

src/main/java/hudson/plugins/jira/JiraCreateIssueNotifier.java
```diff
@@ -411,7 +408,21 @@ private void currentBuildResultSuccess(AbstractBuild<?, ?> build, BuildListener
         }
     }
 
-    private void progressWorkflowAction(AbstractBuild<?, ?> build, String issueId, Integer actionId) throws IOException {
+    static boolean isDone(Status status) {
+        if (status.getName().equalsIgnoreCase(finishedStatuses.Closed.toString()) ||
```
warden: why not only testing the category then?

pom.xml
warden: @olamy I think you had some doubts about guava dependency...

t-8ch: I'll double-check.
The best way forward would probably to focus the client update in #213 and not duplicate the work here.

src/main/java/hudson/plugins/jira/JiraCreateIssueNotifier.java
t-8ch: For backwards compatibility, somebody may have these statuses with different categories.
(Although it would not make sense).
If you are fine with requiring users to have the correct category on their statuses I can drop it.

t-8ch:
Now that #213 has landed I removed the dependency update from this PR.
If you want I can drop the explicitly named statuses from the PR.

src/main/java/hudson/plugins/jira/JiraCreateIssueNotifier.java
warden: Ah, indeed, that makes sense, let's leave it for now.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ See [GitHub releases](https://github.com/jenkinsci/jira-plugin/releases)
 ### Unreleased
 Release date:  _March 2, 2020_
```
olamy: ? we are now using https://github.com/jenkinsci/jira-plugin/releases

warden: @olamy I added CHANGELOG.md before we've started using GH releases, I'll take a look if we could combine both approaches (and I still think it's a good idea to have CHANGELOG directly in git) - we can do both for now I guess.

t-8ch:
@warden Thanks for merging it!
Do you have a rough estimate when the next release including this PR will be cut?

PR #220, opened 2020-04-02
Enable dependabot

warden:
Based on: https://github.com/jenkinsci/plugin-pom/blob/master/.dependabot/config.yml
You can see the trial runs on: https://github.com/warden/jira-plugin/pulls

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -1,0 +1,26 @@
+version: 1
+update_configs:
+- package_manager: "java:maven"
+  directory: "/"
+  update_schedule: "monthly"
+
+  default_reviewers:
+    - "warden"
+    - "olamy"
+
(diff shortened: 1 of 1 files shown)


olamy:
I'm 0 with that. But please do not apply automatically every update proposal. I can see in your fork upgrade such guava. This can be **REALLY**  dangerous to upgrade this in the context of Jenkins
So please **DO NOT ACCEPT ALL PRs FROM THIS BOT BLINDLY**

warden:
Sure, we are the reviewers and automatic merges are off, so it only depends on us if we think it's safe or not. Let's just agree it needs approval of all maintainers for now to get it merged. 
We can also add guava to the exclusions list if needed.

warden:
@olamy , @artkoshelev can you approve please?

PR #222, opened 2020-04-04
Remove all deprecated fields and classes for 4.0

warden:


src/main/java/hudson/plugins/jira/JiraProjectProperty.java
```diff
@@ -83,66 +81,5 @@ public boolean isApplicable(Class<? extends Job> jobType) {
...
-         *
-         * @return array of sites
-         */
-        @Deprecated
-        public JiraSite[] getSites() {
-            return JiraGlobalConfiguration.get().getSites().toArray(new JiraSite[0]);
-        }
-
-        @SuppressWarnings("unused") // Used by stapler
-        public ListBoxModel doFillSiteNameItems(@AncestorInPath AbstractFolder<?> folder) {
```
warden, line 108: bring this back

warden:
@olamy, @artkoshelev PTAL

src/main/java/hudson/plugins/jira/JiraSite.java
```diff
@@ -1085,6 +991,7 @@ public String getDisplayName() {
             return "Jira Site";
         }
 
+        private JiraSession session;
```
artkoshelev, line 994: why was it moved outside of method below?

src/main/java/hudson/plugins/jira/JiraSite.java
```diff
@@ -1172,15 +1081,14 @@ public FormValidation doValidate(@QueryParameter String url,
...
             site.setThreadExecutorNumber(threadExecutorNumber);
-            JiraSession session = null;
+
+            session = null;
             try {
                 session = site.getSession();
                 session.getMyPermissions();
                 return FormValidation.ok("Success");
-            } catch (RestClientException e) {
+            } catch (RuntimeException e) {
```
artkoshelev, line 1091: that's significant change of method contract, seems not related to described changes

src/test/java/hudson/plugins/jira/DescriptorImplTest.java
```diff
@@ -22,12 +24,15 @@
...
 
 import java.io.IOException;
+import java.net.URL;
 import java.util.Arrays;
 import java.util.Iterator;
 
 import static org.hamcrest.Matchers.empty;
 import static org.hamcrest.Matchers.hasSize;
 import static org.junit.Assert.*;
+import static org.mockito.Mockito.*;
```
artkoshelev, line 34: pls avoid wildcard imports

src/main/java/hudson/plugins/jira/JiraSite.java
warden: This change was done to cover the frequent NPE case and to log it accordingly instead of throwing it as this method is supposed to return FormValidation error, so for me the contract stays the same.

warden:
@olamy any thoughts or things you are not happy about? 🤖

PR #223, opened 2020-04-06
DO NOT MERGE fail test on purpose

olamy:

### DIFF
--- a/src/test/java/hudson/plugins/jira/CliParameterTest.java
+++ b/src/test/java/hudson/plugins/jira/CliParameterTest.java
@@ -52,4 +52,10 @@ public class CliParameterTest {
         CLICommandInvoker.Result result = invoker.invokeWithArgs(project.getName(), "-p", "jiraversion=1.0");
         assertThat(result, succeeded());
     }
+
+    @Test
+    public void failing() {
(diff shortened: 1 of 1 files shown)


olamy:
LOL `Travis CI Build passed`

warden:
@olamy yes, that's correct and not a surprise that Travis build passed.
Currently tests are disabled on Travis:
https://github.com/jenkinsci/jira-plugin/blob/master/.travis.yml#L19
Travis is only used currently for SonarCloud/Coveralls results.

I'm fine with moving SC/Cobertura to Jenkins but I'm not super eager to do it as Jenkins feedback are really slow comparing to Travis and also sometimes unstable. (even yesterday I had a build failure because of agent allocation).
Another reason is that I can't re-trigger the build on Jenkins, while I can on Travis.

warden:
I saw today another unexpected failure of Jenkins Pipelines (see #229) and since:
- I don't see a method to properly encrypt credentials for sonar and our Jenkins infra
- configuration-as-code plugin is also using this kind of integration
and in order not to confuse contributors, I decided to reenable full build with tests on Travis to have parity with main Jenkins build.

PR #232, opened 2020-04-23
Add code style checks

warden:
This PR introduces auto-formatting according to [Google JavaGuide Styleguide](https://google.github.io/styleguide/javaguide.html). 
This ensures that all PR coming from the contributors will follow the same codestyle - if they at least compile it that is :-)


warden:
SG/coverage report is irrelevant here as this is reformatting code only.

olamy:
Some prefers blue some prefers green This is always a complicated discussion....
That's why the default consensus on Jenkins is use the style already applied to the current file you are editing or apply the style you like when creating a new file.
I'm definitely not a fan of this style as I prefer this one http://maven.apache.org/developers/conventions/code.html.
By the way I'm -1 using a plugin changing the sources per default. This can generate so many bugs...... (maybe in a profile)
Configure m-checkstyle-p to fail the build if formatting is not correct and this will teach the developper to apply  the correct style....

artkoshelev:
+1 for unified style for the project
-1 for auto-formatting: as @olamy said, let's configure m-checkstyle-p to fail the build if formatting is not correct

warden:
I'm OK with checkstyle as well - honestly I don't care as long as all the code gets checked against the same rules.
Does either of you already have the rules/configuration for checkstyle and can commit it in or should I just go with the default config?

warden:
I used the same rules that [configuration-as-code](https://github.com/jenkinsci/configuration-as-code-plugin) is using. The build obviously fails right now as there are many fixes to be done, but I want us to agree that we're ok with those rules before I start fixing them, let me know!

artkoshelev:
LGTM

warden:
1. No, the `.idea/codeStyle` directory is there on purpose to automatically configure the codestyle according to `.mvn` directory. This is for automation so that nobody needs to do it manually, [this is a common approach](https://github.com/jenkinsci/configuration-as-code-plugin/tree/master/.idea/codeStyles).

2. For the style, I'm all ears and I can change it, but we need to agree on a style that will work with everyone. If there is no proposal, let's just go with what other plugins are using.

olamy:
related to https://plugins.jetbrains.com/plugin/1065-checkstyle-idea?
It's only a matter of using the checkstyle idea plugin to use the checkstyle file.
If we change the checkstyle file because it's used by maven plugin we will have to regenerate the proprietary idea files so the risk about not being sync... That's why I don't like the idea about those idea proprietary files. if you prefer it use it locally it doesn't hurt.
used in one place doesn't mean it's common to use it :)

.idea/checkstyle-idea.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="CheckStyle-IDEA">
```
warden, line 3: @olamy yes, that is the CheckStyle-IDEA plugin that I'm using.
The file is using relative paths so the risk of being out-of-sync should be minimal, but I'm not going to make a fuss about it and I can remove it and instead add a README entry about it.

warden:
The more important part is 2) - what about the code style then, any other proposals?

olamy:
I don't like much this style but I don't mind as long as there is the checkstyle file.
I prefer Jetty style https://github.com/eclipse/jetty.project/blob/jetty-10.0.x/build-resources/src/main/resources/jetty-checkstyle.xml or Maven https://github.com/apache/maven-shared-resources/blob/master/src/main/resources/config/maven_checks.xml

.idea/checkstyle-idea.xml
warden: I just found [this](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839-How-to-manage-projects-under-Version-Control-Systems), so this is the recommended way to share IDEA configuration between developers.

This is project-specific so it should hurt to have it there and I think it will make contributors life easier, because the project will be auto-configured .

warden:
@artkoshelev do you have any preference between those 3?

artkoshelev:
i do prefer google style

.idea/checkstyle-idea.xml
warden: @olamy do you want me to revent all changes in .idea directory and not commit it there?
I can add a README entry for code formatting setup instead or we can allow this in the repo as explained in the link above.

warden:
I looked at the google_checks.xml and we're far away from having this, so I only added basic indentation stuff from there and I propose to start small.
The PR is splitted into two distinguishable commits for easier reference.

olamy:
https://twitter.com/kohsukekawa/status/150279633055657985 :)

olamy:
@warden what I mean is can we move forward because those changes will screw up every PR and every forks so if we do it we have to do it quickly otherwise everybody is blocked ...

olamy:
Are you sure about the file `.mvn/checkstyle.xml`? It doesn't look to be the google java style so the Intellij files seems different? Please look here https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/google_checks.xml
Again I don't want we depend on IDE specific files. What if Eclipse, Netbeans... users want to contribute

warden:
Yes I am sure about the checkstyle file and no, it is not google_style - please read earlier comments. That file defines the rules and it is used by the checkstyle plugin, so it is not related to IDE.

warden:
`idea/*` -> https://github.com/jenkinsci/configuration-as-code-plugin/commit/20062781d91800c645fa0948a8f99d355c35b880
If you suggest to revert changes in .gitignore and .idea, let me know - I'm waiting for constructive proposal to move forward.

olamy:
if we do checkstyle let's do for real and use google style or any style used by another opensource project.
@artkoshelev mentioned he was ok for google style
I'm ok to use checkstyle but please a standard one and not a new homemade style (we are not here to reinvent the wheel). 
regarding `.idea/checkstyle-idea.xml` I see 
```
        <entry key="location-0" value="BUNDLED:(bundled):Sun Checks" />
        <entry key="location-1" value="BUNDLED:(bundled):Google Checks" />
        <entry key="location-2" value="PROJECT_RELATIVE:$PROJECT_DIR$/.mvn/checkstyle.xml:.mvn/checkstyle.xml" />
```

Is it a preference order?

.idea/checkstyle-idea.xml
```diff
@@ -0,0 +1,18 @@
...
+    <option name="configuration">
+      <map>
+        <entry key="active-configuration" value="PROJECT_RELATIVE:$PROJECT_DIR$/.mvn/checkstyle.xml:.mvn/checkstyle.xml" />
+        <entry key="checkstyle-version" value="8.22" />
+        <entry key="copy-libs" value="false" />
+        <entry key="location-0" value="BUNDLED:(bundled):Sun Checks" />
```
olamy, line 9: is it an order preference?

.idea/checkstyle-idea.xml
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="CheckStyle-IDEA">
+    <option name="configuration">
+      <map>
+        <entry key="active-configuration" value="PROJECT_RELATIVE:$PROJECT_DIR$/.mvn/checkstyle.xml:.mvn/checkstyle.xml" />
+        <entry key="checkstyle-version" value="8.22" />
```
olamy, line 7: how are we sure this version is in sync with maven-checkstyle-plugin?

.idea/checkstyle-idea.xml
warden: No, see the key `active-configuration`. Those 2 locations (0,1) are bundled and cannot be removed.

.idea/codeStyles/Project.xml
```diff
@@ -0,0 +1,128 @@
...
+        <option name="CONTINUATION_INDENT_SIZE" value="4" />
+      </indentOptions>
+    </codeStyleSettings>
+    <codeStyleSettings language="JSON">
+      <indentOptions>
+        <option name="CONTINUATION_INDENT_SIZE" value="4" />
+        <option name="TAB_SIZE" value="2" />
+      </indentOptions>
+    </codeStyleSettings>
+    <codeStyleSettings language="JavaScript">
```
olamy, line 98: is it related to this project? I don't think we have javascript files.

.idea/codeStyles/Project.xml
```diff
@@ -0,0 +1,128 @@
...
+      <option name="IF_BRACE_FORCE" value="3" />
+      <option name="DOWHILE_BRACE_FORCE" value="3" />
+      <option name="WHILE_BRACE_FORCE" value="3" />
+      <option name="FOR_BRACE_FORCE" value="3" />
+      <indentOptions>
+        <option name="INDENT_SIZE" value="2" />
+        <option name="CONTINUATION_INDENT_SIZE" value="4" />
+      </indentOptions>
+    </codeStyleSettings>
+    <codeStyleSettings language="JSON">
```
olamy, line 92: is it related to this project?

warden:
[As I mentioned earlier](https://github.com/jenkinsci/jira-plugin/pull/232#issuecomment-621341450), we cannot use Google Style fully as it contains a lot of rules that the current code does not comply with. For example, it has a rule that a method must have JavaDoc.
That's why I decided to start small and only apply code-formatting rules + some basic stuff like wildcard import checks.

Alternative is to take google_checks.xml and customise the threshold to warning for the above-mentioned issues.

.idea/codeStyles/Project.xml
warden: This is how codeStyle saves configuration.

.idea/codeStyles/Project.xml
warden: This is how codeStyle saves configuration.

.idea/checkstyle-idea.xml
warden: I was wondering about it as well but I didn't find an alternative - I'm ok with removing this file altogether and adding a README entry how to set up checkstyle.

PR #233, opened 2020-04-30
Add Google codestyle rules

warden:
This PR adds Google Java styleguide settings-only (for IDEA) and updates the README about the code style guide. It does not run any enforced checks nor does not contain any reformatting, but serves as a first step to improve gradually the codebase.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,8 +5,13 @@
 build
 target
 bin
+
+# ignore all Idea files except for common codestyle settings
 *.iml
-.idea/
+.idea/*
+!.idea/checkstyle-idea.xml
+!.idea/google-java-format.xml
+
 work
 atlassian-ide-plugin.xml
 .work/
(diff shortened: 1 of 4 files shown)


warden:
This is a light alternative to #232 to keep @olamy but make some progress :-)

warden:
I don't see any nays so far, so I'm going to merge it after the weekend.

.gitignore
```diff
@@ -5,8 +5,14 @@
 build
 target
 bin
+
+# ignore all Idea files except for common codestyle settings
 *.iml
-.idea/
+.idea/*
+!.idea/checkstyle-idea.xml
+!.idea/codeStyles/
```
olamy, line 13: is it style correct?

warden: What do you mean? this is not a .java file.

olamy: as long as you have this directory with files it will not be ignored by git. AFICS we do not add anything from this directory `.idea/codeStyles/` we have only 2 files to not ignore `.idea/checkstyle-idea.xml ` and `.idea/google-java-format.xml`

warden: Ah, you meant this line is extra, good catch! leftover from the other approach.

warden: Fixed!

PR #239, opened 2020-05-07
simplify pom, remove non used dependencies and use bom for LTS we use. need some changes in mockito

olamy:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -19,10 +19,8 @@
     <revision>3.0.16</revision>
     <changelist>-SNAPSHOT</changelist>
     <java.level>8</java.level>
-    <mockito.version>1.10.19</mockito.version>
     <jira-rest-client.version>5.2.1</jira-rest-client.version>
     <fugue.version>3.0.0</fugue.version>
-    <slf4jVersion>1.7.26</slf4jVersion>
     <!-- jenkins -->
     <jenkins.version>2.176.4</jenkins.version>
     <!-- security -->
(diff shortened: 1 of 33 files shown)


olamy:
still in draft as using mockito core version from jenkins bom need some changes. But this should fix PCT failures with jdk11

warden:
I actually did similar changes locally in my feature PR but got stuck in on the Mockito update with the unexpected number of issues to be fixed with it,  so that would definitely help!

pom.xml
```diff
@@ -119,6 +111,18 @@
...
        </executions>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <!-- not reuse because Mockito record stuff statically and some are changing between tests so it suc... -->
+          <reuseForks>false</reuseForks>
+          <!-- to run some tests within your IDE you will need this -->
+<!--          <systemPropertyVariables>-->
```
warden, line 121: Oh, now it's clear - how about adding this to a profile?

src/main/java/hudson/plugins/jira/JiraJobAction.java
```diff
@@ -31,7 +32,7 @@
  * e.g. "JENKINS-1234" and "feature/JENKINS-1234" will have this action with the issue JENKINS-1234 referenced
  */
 @ExportedBean
-public class JiraJobAction implements Action {
+public class JiraJobAction implements Action, Serializable {
```
warden, line 35: Why is this required

olamy: good catch sorry I miss to remove this. done

warden: Please organise imports - https://sonarcloud.io/project/issues?id=jenkins-jira-plugin&open=AXIIy9Pg99-5isQx0vpU&pullRequest=239&resolved=false&types=CODE_SMELL

warden:
LGTM, ready for squash merge?

PR #240, opened 2020-05-22
add possibility to use environment variables in description

Simmay93:


warden:
Thanks for contributing to the plugin!
Can you review the[ contibution guidelines and add a test](https://github.com/jenkinsci/jira-plugin#contributing-to-the-plugin)?

Simmay93:
Unfortunately I cannot do that. I don't have the infrastructure and know-how to build this project. Furthermore I have not worked yet with Mockito. So I don't know how to implement the test.

In my opinion there should be added a test in "JiraCreateIssueNotifierTest.java" in which an evironment variable is set and extended in the description. Should not be that complicated.

I can try it myself but then I'd be happy to have some help to initially build the project. I did not find instructions and requirements in the README.

warden:
You have modified the source code and created a PR, so I think it's safe to assume you did that because you built a custom SNAPSHOT plugin package with `mvn clean package` and deployed it to your local Jenkins instance, am I correct?

If that's the case, then you have everything you need - you can run the tests with standard `mvn clean test` command. For Mockito, check the already existing tests that are in `src/test/` directory. Also take a look at [this](https://github.com/jenkinsci/jira-plugin/blob/82fcceae7808bf1f98542defc00d6a313bb9954e/src/test/java/hudson/plugins/jira/VersionCreatorTest.java#L70) similar test case.

Simmay93:
Wasn't that easy but learned a lot. ;-)

I added a test case and also adopted an existing test case to have more coverage.

warden:
There's a small issue with the build that is not related to this PR, we'll fix it in the meantime, check if all tests are passing and merge it then.
Glad that the mentioned resources helped, thanks for contribution!

Simmay93:
In my local Jenkins environment it works as expected. Is there anything else I should do?

warden:
All should be good from your side, please wait a while we resolve the build issue and then this will be merged.

PR #242, opened 2020-05-26
Avoid failure if there is malformed versions in the list

gcuisinier:
We wanted to used the JIRa plugin in our Jenkins to add JIRA release Parameter, but we have project that contains (but used) strange name as version :

"201601-"
"COMPONENT-"

Due to the present of "-", the plugin try to parse it and expect something after the "-".
But in our case, it's not the case.
And it's failing, and that causes the list of the parameter to be empty.

Here is a small fix that avoid to throw exception in this case and allow to return the list of release.

### DIFF
--- a/src/main/java/hudson/plugins/jira/versionparameter/VersionComparator.java
+++ b/src/main/java/hudson/plugins/jira/versionparameter/VersionComparator.java
@@ -41,7 +41,7 @@ public class VersionComparator implements Comparator<Version> {
      */
     protected String getNumberVersion(String firstV) {
         String res = firstV;
-        if (!firstV.matches("[0-9.]+") && firstV.contains("-")) {
+        if (firstV.contains("-")) {
             String[] splittedVersion = firstV.split("-");
             if(splittedVersion.length > 1) {
                 res = splittedVersion[1];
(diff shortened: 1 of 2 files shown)


gcuisinier:
Moreover, in our case, the release is an old release, I hoped that not showing Released version can help.

But as the sort is done before the filtering, it's not working 

https://github.com/jenkinsci/jira-plugin/blob/master/src/main/java/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinition.java#L74

src/main/java/hudson/plugins/jira/versionparameter/VersionComparator.java
```diff
@@ -42,7 +42,10 @@ public int compare(Version rev1, Version rev2) {
     protected String getNumberVersion(String firstV) {
         String res = firstV;
         if (!firstV.matches("[0-9.]+") && firstV.contains("-")) {
```
artkoshelev, line 44: can you pls also remove number regexp matching here since it will never match if there is a '-' in the input string anyway?

gcuisinier: Indeed. I didn't want modify the existing condition, but you're right.

Adding commit is ok ? Or do you want me to rebase?

gcuisinier:
Thank you :)

Any idea of the date for the next release ? Or not at all ? :)

olamy:
no worries. early next week sounds good for a release.

PR #248, opened 2020-06-03
Bump Jenkins requirement to 2.204 from plugin-pom

warden:


warden:
Needed to see build results from Jenkins - I'll investigate locally.

PR #273, opened 2020-09-30
Suppressing JiraBuildActionTest.binaryCompatibility which misuses XStream

jglick:
Compatibility with https://github.com/jenkinsci/jenkins/pull/4944 in PCT. After https://github.com/jenkinsci/jenkins/pull/4944/commits/0e5aeec4ccc67207c0912ffe471fe5cf2815cf04, the mock of `FreeStyleBuild` has a `writeReplace` method which returns null, which is illegal and breaks things:

```
java.lang.RuntimeException: Failed to serialize hudson.plugins.jira.deprecated.DeprecatedJiraBuildAction#owner for class hudson.plugins.jira.deprecated.DeprecatedJiraBuildAction
	at hudson.util.RobustReflectionConverter$2.writeField(RobustReflectionConverter.java:255)
	at hudson.util.RobustReflectionConverter$2.visit(RobustReflectionConverter.java:223)
	at com.thoughtworks.xstream.converters.reflection.PureJavaReflectionProvider.visitSerializableFields(PureJavaReflectionProvider.java:150)
	at hudson.util.RobustReflectionConverter.doMarshal(RobustReflectionConverter.java:209)
	at hudson.util.RobustReflectionConverter.marshal(RobustReflectionConverter.java:150)
	at com.thoughtworks.xstream.core.AbstractReferenceMarshaller.convert(AbstractReferenceMarshaller.java:68)
	at com.thoughtworks.xstream.core.TreeMarshaller.convertAnother(TreeMarshaller.java:58)
	at com.thoughtworks.xstream.core.TreeMarshaller.convertAnother(TreeMarshaller.java:43)
	at com.thoughtworks.xstream.core.TreeMarshaller.start(TreeMarshaller.java:82)
	at com.thoughtworks.xstream.core.AbstractTreeMarshallingStrategy.marshal(AbstractTreeMarshallingStrategy.java:37)
	at com.thoughtworks.xstream.XStream.marshal(XStream.java:1250)
	at com.thoughtworks.xstream.XStream.marshal(XStream.java:1239)
	at com.thoughtworks.xstream.XStream.toXML(XStream.java:1212)
	at com.thoughtworks.xstream.XStream.toXML(XStream.java:1199)
	at hudson.plugins.jira.JiraBuildActionTest.binaryCompatibility(JiraBuildActionTest.java:29)
	at …
Caused by: java.lang.NullPointerException
	at hudson.util.RobustReflectionConverter.marshal(RobustReflectionConverter.java:143)
	at com.thoughtworks.xstream.core.AbstractReferenceMarshaller.convert(AbstractReferenceMarshaller.java:68)
	at com.thoughtworks.xstream.core.TreeMarshaller.convertAnother(TreeMarshaller.java:58)
	at com.thoughtworks.xstream.core.AbstractReferenceMarshaller$1.convertAnother(AbstractReferenceMarshaller.java:83)
	at hudson.util.RobustReflectionConverter.marshallField(RobustReflectionConverter.java:264)
	at hudson.util.RobustReflectionConverter$2.writeField(RobustReflectionConverter.java:251)
	... 42 more
```

Amends #72.

### DIFF
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -2,6 +2,6 @@
   <extension>
     <groupId>io.jenkins.tools.incrementals</groupId>
     <artifactId>git-changelist-maven-extension</artifactId>
-    <version>1.0-beta-7</version>
+    <version>1.2</version>
   </extension>
 </extensions>
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>4.7</version>
+    <version>4.8</version>
     <relativePath />
   </parent>
 
@@ -18,6 +18,7 @@
   <properties>
     <revision>3.1.2</revision>
     <changelist>-SNAPSHOT</changelist>
+    <gitHubRepo>jenkinsci/jira-plugin</gitHubRepo>
     <java.level>8</java.level>
     <jira-rest-client.version>5.2.1</jira-rest-client.version>
     <fugue.version>3.0.0</fugue.version>
@@ -28,10 +29,10 @@
   </properties>
 
   <scm>
-    <connection>scm:git:git://github.com/jenkinsci/jira-plugin.git</connection>
-    <developerConnection>scm:git:git@github.com:jenkinsci/jira-plugin.git</developerConnection>
-    <url>https://github.com/jenkinsci/jira-plugin</url>
-    <tag>jira-3.1.0</tag>
+    <connection>scm:git:git://github.com/${gitHubRepo}.git</connection>
+    <developerConnection>scm:git:git@github.com:${gitHubRepo}.git</developerConnection>
+    <url>https://github.com/${gitHubRepo}</url>
+    <tag>HEAD</tag>
   </scm>
 
   <licenses>
--- a/pom.xml
+++ b/pom.xml
@@ -32,7 +32,7 @@
     <connection>scm:git:git://github.com/${gitHubRepo}.git</connection>
     <developerConnection>scm:git:git@github.com:${gitHubRepo}.git</developerConnection>
     <url>https://github.com/${gitHubRepo}</url>
-    <tag>HEAD</tag>
+    <tag>${scmTag}</tag>
   </scm>
 
   <licenses>
--- a/src/test/java/hudson/plugins/jira/JiraBuildActionTest.java
+++ b/src/test/java/hudson/plugins/jira/JiraBuildActionTest.java
@@ -10,6 +10,7 @@ import org.junit.Test;
 import hudson.model.FreeStyleBuild;
 import hudson.plugins.jira.deprecated.DeprecatedJiraBuildAction;
 import hudson.util.XStream2;
+import org.junit.Ignore;
 
 /**
  * Test if existing serialized JiraBuildAction information
@@ -19,6 +20,7 @@ import hudson.util.XStream2;
  */
 public class JiraBuildActionTest {
 
+    @Ignore("TODO do not use mock frameworks for this; use @LocalData")
     @Test
     public void binaryCompatibility() {
         XStream2 xStream2 = new XStream2();


jglick:
> the tests on `master` are now passing

Yes, but they will not against a future version of Jenkins (possibly weekly releases in about a month, after the next LTS line is branched). It is standard practice to run `plugin-compat-tester` against commonly used plugins plus newer versions of Jenkins core, and spurious failures are a drag on productivity. Fixing tests proactively keeps things running smoothly.

> actually fix the test

This would be great, for someone intimately familiar with the plugin’s functionality and the original intent behind the test case, whatever that was. In the meantime, an ignored test is better than a broken test.

> according to new API/standards

These are _not_ new standards (nor is it a new API, really). Mockito/Powermock are well known to cause extremely fragile test cases that stop working when only tangentially related aspects of implementation code are modified. This one just happens to be breaking now. The correct way to test deserialization compatibility of plugin classes is to use `@LocalData`, as mentioned in the comment, which along with `JenkinsRule` exercises the actual code paths Jenkins would use when loading old `config.xml` files from disk.

warden:
> Yes, but they will not against a future version of Jenkins (possibly weekly releases in about a month, after the next LTS line is branched). It is standard practice to run `plugin-compat-tester` against commonly used plugins plus newer versions of Jenkins core, and spurious failures are a drag on productivity. Fixing tests proactively keeps things running smoothly.

OK, so this is the initial hidden reason for this PR. @jglick please bear with me while I'm trying to decompose this by splitting this into two relevant issues:

> spurious failures are a drag on productivity. Fixing tests proactively keeps things running smoothly.

Yes of course, we all agree with that - that's why the test should be fixed and not ignored, in my view it should stay as it is and fail only in the moment when someone tries to use the newer Jenkins core as direct dependency, but...
on the other side:

>  It is standard practice to run `plugin-compat-tester` against commonly used plugins plus newer versions of Jenkins core

So, now as I understand here is the actual issue and that's a 'workflow problem'.
I think that if the PCT tests test the compatibility _of a plugin_ then they should run on the pipelines _of the plugin_ and not somewhere else - in order to provide early feedback to the actual plugin developers/maintainers, what do you think about this approach?

And in such case as this, even though the PCT pipeline detected a problem, I believe it would be a better practice to "flag" the problem, by a draft PR or an auto-opened GH issue - instead of just ignoring the test that (if I haven't missed anything) wouldn't bring any value other that not failing the PCT pipeline "somewhere".
(I wrote somewhere here, as there is no direct visibility on any of the pipelines in `jira-plugin` CI space - or I'm unaware of it)

The least that I could do from a maintainers perspective is to add a PCT information section in our plugin's README as a reference.

- is there anything I have missed that would "break" if we don't merge it and would leave it as a DRAFT?
- would it be anyhow feasible to move the PCT tests into our space for direct visibility?

jglick:
Sorry for not setting more context up front. https://github.com/jenkinsci/jep/pull/309 may be helpful. Basically I am trying to logically prepare a patch to all of Jenkins, core and plugins, to a state where we use the official version of XStream, and looking for any new test failures. My employer distributes an OEM copy of Jenkins which is based on a slightly tweaked LTS release plus a broad set of plugins, of which `jira` is one, in specific versions. We have internal tooling which runs `acceptance-test-harness` and `plugin-compat-tester` on this whole aggregate, with whatever proposed patches, and gathers tens of thousands of test results. This test showed up as a regression, and in fact I can reproduce that failure merely by setting `-Djenkins.version=…-SNAPSHOT`; diagnosis shows that the failure is the fault of the test, not production code.

> should run on the pipelines _of the plugin_

Yes, this has certainly been proposed. The question remains what, exactly, the tested software versions should be. Last weekly release of Jenkins? Fine, but then a problem like this would only become visible _after_ merging the core PR, which is unsettling. (And making the plugin’s CI test against an ever-changing version would mean that trivial PRs editing `README.md` could blow up with test failures for reasons totally unrelated to the code change, among other problems; so you would need to pin a weekly version number and then periodically bump it with Dependabot, or something.)

> is there anything I have missed that would "break" if we don't merge it and would leave it as a DRAFT?

“Break”? No. It would make my life slightly more difficult, by having to maintain this fork of the plugin for integration testing for weeks or months. Fortunately https://jenkins.io/jep/305 means that when ci.jenkins.io is happy with my patch, I can refer to an exact fork commit in the manifest of core and all plugins touched by the XStream change; tests which rely on binaries (like ATH) will use the deployed `*.hpi`, and tests which run on sources (PCT) will use the commit from my fork.

> the test should be fixed and not ignored

In general yes. In this case I suspect the value of the test was always very low. Mockito-based tests are in my experience not only a maintenance burden, but frankly close to useless in terms of detecting genuine regressions: the code path being exercised is so synthetic that it is unlikely to overlap with the sorts of behavioral changes in upstream components (such as Jenkins core) that tend to cause surprising compatibility errors.

About the only thing the current test actually verifies is the behavior of the `owner` field, which was wrong before #72 and is still wrong: this should be `transient`, via `RunAction2`.

warden:
Apart from the rest of discussion - I'll take a look to fix the test as you suggested so assigning to me, as not to merely ignore the test, so that one way or the other I'll address your scenario. 
Many thanks for the great context write-up and your thoughts about Mockito tests!

> The question remains what, exactly, the tested software versions should be. Last weekly release of Jenkins? Fine, but then a problem like this would only become visible after merging the core PR, which is unsettling

Why is it unsettling?
The current scenario where external projects like yours are using PCT tests to discover potential incompatibilities is good, provided those "downstream" projects provide actual fixes to the code, to address both scenarios:
- keep compatibility with the upstream (here jenkins/core)
- satisfy own usecases (here your downstream project)
- do not break any functionality or tests inside the project itself

So overally, that's great that you discovered this and that you are running those tests in your pipelines (those are your CI resources anyway :-) )

Just trying to figure out a way to proactively include it in our own plugin's CI system to test.

> The question remains what, exactly, the tested software versions should be. Last weekly release of Jenkins? Fine, but then a problem like this would only become visible after merging the core PR, which is unsettling.

Last released version should be fine I think. The part " a problem like this would only become visible after merging the core PR" is correct and imo that's how it should work. The upstream introduced a change and only after it lands and is released, the plugin maintainers should worry about it.

Another approach would be to do "release management" kind of work, where we would label all affected plugins with some core change and only release when all of those are fixed - but this is in a commercially supported (aka paid for someone's time to actually fix stuff) projects rather than here I think.

jglick:
> I'll take a look to fix the test

Great. Basically just need to have a pre-#72 build record in a `src/test/resources/hudson/plugins/jira/JiraBuildActionTest/binaryCompatibility/jobs/xxx/build.xml` along with the `…/jobs/xxx/config.xml`, mark with `@LocalData`, then assert that after you load that build (`Jenkins.getItemByFullName`, `getBuildByNumber`) that you can `.getAction(JiraBuildActionTest.class)` and it has the right stuff in it.

Then you can actually fix the mistake in production code by making `owner` a `transient` field initialized by implementing `RunAction2`.

jglick:
> Just trying to figure out a way to proactively include it in our own plugin's CI system to test.

Well, you can already specify some alternative version of Jenkins core to test against. (We used to try to do this automatically but caused a lot of problems; see https://github.com/jenkins-infra/pipeline-library/pull/145 for discussion.) Note that this is not quite as strong as PCT, which also tests against updated _plugin_ dependencies, though you can get a similar effect by just using Dependabot (which is really only practical if you also use the [BOM](https://github.com/jenkinsci/bom#usage)).

A topic for the dev list.

warden:
I followed your hints but it seems I am missing something, XML shows the Deprecated class. Isn't it because in the old class we have Build in the constructor and in the new one we have Run?
https://github.com/jenkinsci/jira-plugin/commit/7c0d18585f9242c9ed7b5bef0819f84cdd1f8a29

jglick:
BTW you can remove `owner` from the `JiraBuildAction` constructor; `onAttached` handles that.

jglick:
> XML shows the Deprecated class.

Not sure what you mean. The old `build.xml` will keep the `owner` field unless and until the `Run` is `save`d.

jglick:
Also you likely want to assert that the `owner` field is indeed set.

jglick:
And you should not need `JenkinsConfiguredWithCodeRule`. Sorry for the awkward commenting, but unless you file a PR for your changes, I cannot comment directly.

jglick:
(Note that if you do merge #283 using the simple history-aware merge strategy, rather than squash or rebase, then that would automatically mark this PR as merged as well: its head commit would become an ancestor of the new head of its base branch.)

PR #282, opened 2020-10-02
Re-add ignores to dependabot

warden:


olamy (approved):
yup sorry I missed that

PR #283, opened 2020-10-02
Fix JiraBuildAction test and logic

warden:
* use RunAction2 for initialisation of action
* stop comparing XML in tests in favor of full-workflow object comparison after deserialization

Replaces #273


src/test/java/hudson/plugins/jira/JiraBuildActionTest.java
```diff
@@ -1,38 +1,47 @@
...
+
+    Job job = r.getInstance().getItemByFullName("/project", Job.class);
+    Run run = job.getBuildByNumber(2);
+    assertEquals("job/project/2/", run.getUrl());
+
+    XStream2 xStream2 = new XStream2();
+    String xml = xStream2.toXML(run);
+    System.out.println(xml);
+
+    JiraBuildAction jba = run.getAction(JiraBuildAction.class);
```
warden, line 43: this throws NPE, that's why I was looking at xml of `run`

src/main/java/hudson/plugins/jira/JiraBuildAction.java
```diff
@@ -1,73 +1,84 @@
...
+  public transient Run<?, ?> owner;
 
-    private Set<JiraIssue> issues;
+  private Set<JiraIssue> issues;
 
-    public JiraBuildAction(@Nonnull Run<?, ?> owner, @Nonnull Set<JiraIssue> issues) {
-        this.owner = owner;
-        this.issues = new HashSet( issues);
-    }
+  public JiraBuildAction(@Nonnull Run<?, ?> owner, @Nonnull Set<JiraIssue> issues) {
```
jglick: or drop `owner` parameter and let `onAttached` set the field

jglick:
BTW https://github.com/jenkinsci/jira-plugin/pull/283/commits/b4f3d727bd3f2a92ddb8b38075978747b0f5472f makes this harder to review…

src/test/resources/hudson/plugins/jira/JiraBuildActionTest/binaryCompatibility/jobs/project/builds/2/build.xml
```diff
@@ -0,0 +1,38 @@
...
+          <hudson.model.Cause_-UserIdCause>
+            <userId>admin</userId>
+          </hudson.model.Cause_-UserIdCause>
+          <int>2</int>
+        </entry>
+      </causeBag>
+    </hudson.model.CauseAction>
+
+    <hudson.plugins.jira.deprecated.DeprecatedJiraBuildAction>
+      <owner class="build" reference="../../.."/>
```
jglick, line 28: FYI, this weird syntax is the best you can hope for if you do not use `RunAction2`. On occasion some race condition actually causes the field to serialize the entire content of the `Run`, leading to very colorful errors.

src/test/java/hudson/plugins/jira/JiraBuildActionTest.java
jglick: Throws NPE from what?

warden:
> BTW [b4f3d72](https://github.com/jenkinsci/jira-plugin/commit/b4f3d727bd3f2a92ddb8b38075978747b0f5472f) makes this harder to review…

Just removed reformatting commit for now.

src/main/java/hudson/plugins/jira/JiraBuildAction.java
```diff
@@ -18,12 +19,22 @@
...
+    public void onAttached(Run<?, ?> r) {
+        this.owner = r;
+    }
+
+    @Override
+    public void onLoad(Run<?, ?> r) {
+        this.owner = r;
+    }
+
     public JiraBuildAction(@Nonnull Run<?, ?> owner, @Nonnull Set<JiraIssue> issues) {
```
jglick, line 38: can drop `owner`

src/test/java/hudson/plugins/jira/JiraBuildActionTest.java
```diff
@@ -1,38 +1,59 @@
...
-        String xml = xStream2.toXML(deprecatedJiraBuildAction);
-        System.out.println(xml);
-        xml = xml.replaceAll("hudson.plugins.jira.deprecated.DeprecatedJiraBuildAction", "hudson.plugins.jira.JiraBuildAction");
-        Object fromXML = xStream2.fromXML(xml);
-        System.out.println(fromXML.getClass().getName());
-        JiraBuildAction jiraBuildAction = (JiraBuildAction) fromXML;
-        System.out.println(jiraBuildAction.getDisplayName());
-    }
+  @Rule
+  public JenkinsRule r = new JenkinsConfiguredWithCodeRule();
```
jglick, line 34:
```suggestion
  public JenkinsRule r = new JenkinsRule();
```

jglick (approved):

src/test/java/hudson/plugins/jira/JiraBuildActionTest.java
warden: Just from getAction() method:
```
java.lang.NullPointerException
	at hudson.plugins.jira.JiraBuildActionTest.binaryCompatibility(JiraBuildActionTest.java:55)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.jvnet.hudson.test.JenkinsRule$1.evaluate(JenkinsRule.java:599)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:288)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:282)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.lang.Thread.run(Thread.java:748)
```

jglick: Line number 55 is checking `jba.owner`, are you sure?

warden: Exactly, jba is null in my debug session..

src/test/resources/hudson/plugins/jira/JiraBuildActionTest/binaryCompatibility/jobs/project/builds/2/build.xml
```diff
@@ -0,0 +1,38 @@
...
+        <entry>
+          <hudson.model.Cause_-UserIdCause>
+            <userId>admin</userId>
+          </hudson.model.Cause_-UserIdCause>
+          <int>2</int>
+        </entry>
+      </causeBag>
+    </hudson.model.CauseAction>
+
+    <hudson.plugins.jira.deprecated.DeprecatedJiraBuildAction>
```
jglick, line 27:
```suggestion
    <hudson.plugins.jira.deprecated.JiraBuildAction>
```

and similarly below and delete sources for `DeprecatedJiraBuildAction`, which was just a trick for the old test, not actual production code IIUC.

How did you actually manage to create this `build.xml`?

src/test/java/hudson/plugins/jira/JiraBuildActionTest.java
```diff
@@ -1,38 +1,44 @@
...
 
-import hudson.model.FreeStyleBuild;
-import hudson.plugins.jira.deprecated.DeprecatedJiraBuildAction;
-import hudson.util.XStream2;
 
 /**
- * Test if existing serialized JiraBuildAction information
- * will be loaded after upgrading jira plugin version to with
- * new JiraBuildAction class version introduced in PR-72.
+ * Place needed resources in src/test/resources/compatibility
```
jglick, line 17:
```suggestion
 * Place needed resources in src/test/resources
```

jglick (approved):

jglick (approved):

src/main/java/hudson/plugins/jira/JiraBuildAction.java
```diff
@@ -22,68 +20,68 @@
...
-                return issue;
-            }
-        }
-        return null;
-    }
-
-    public void addIssues(Set<JiraIssue> issuesToBeSaved) {
-        this.issues.addAll(issuesToBeSaved);
+  private static final long serialVersionUID = 457861802965070978L;
+  private final transient Set<JiraIssue> issues;
```
jglick: `transient` means it will never be saved! That is not what you want.

Sonar is just being dumb—it does not notice that the implementation class `HashSet` _is_ `Serializable`.
```suggestion
  private final HashSet<JiraIssue> issues;
```

jglick (changes requested):

src/main/java/hudson/plugins/jira/JiraBuildAction.java
warden: Yeah, I was just looking how to get CI happy temporarilly, but this is also coming from SpotBugs. I'll revert to non-implementing `Serializable` and see how it works then.

jglick: Oh right, there is no need to implement `Serializable` on `JiraBuildAction`; XStream will ignore that. (Same for the `serialVersionUID`.)

src/main/java/hudson/plugins/jira/JiraBuildAction.java
```diff
@@ -1,73 +1,86 @@
...
-import javax.annotation.Nonnull;
-
 /**
  * Jira issues related to the build.
  *
  * @author Kohsuke Kawaguchi
  */
 @ExportedBean
-public class JiraBuildAction implements Action {
+public class JiraBuildAction implements Action, RunAction2 {
```
jglick, line 21:
```suggestion
public class JiraBuildAction implements RunAction2 {
```

(redundant)

src/main/java/hudson/plugins/jira/JiraBuildAction.java
```diff
@@ -1,73 +1,86 @@
...
 
-    public final Run<?, ?> owner;
+  private final HashSet<JiraIssue> issues;
+  private transient Run<?, ?> owner;
 
-    private Set<JiraIssue> issues;
+  public JiraBuildAction(@Nonnull Set<JiraIssue> issues) {
+    this.issues = new HashSet(issues);
```
jglick, line 27:
```suggestion
    this.issues = new HashSet<>(issues);
```

jglick (approved):
:+1:

olamy (approved):

src/main/java/hudson/plugins/jira/JiraBuildAction.java
```diff
@@ -1,73 +1,86 @@
...
  *
  * @author Kohsuke Kawaguchi
  */
 @ExportedBean
-public class JiraBuildAction implements Action {
+public class JiraBuildAction implements RunAction2 {
 
-    public final Run<?, ?> owner;
+  private final HashSet<JiraIssue> issues;
+  private transient Run<?, ?> owner;
```
warden, line 24: @jglick thanks a lot for helping with this PR!
One question, as your comment about `transient` got me thinking.
Sure, we don't want `issues` to be transient in order for them to be serialized into XML, but isn't it the same with `owner`? I mean, where does `RunAction2` know from to which build it should link it to? 
Is it Jenkins that is saving it somewhere internally?

jglick: Yes `Run` knows to inject itself into a `RunAction2` both in `addAction` and upon deserialization. The same applies to various other APIs with “owner” references, like `JobProperty`.

PR #292, opened 2020-10-23
[JENKINS-64002] Fix optional dependencies to other Jenkins plugins

warden:
When installing the plugin on a fresh install without any plugins installed, the *Manage* screen throws ClassNotFound error.

This should be a compilation failure, but since credentials plugin was coming from optional p4 dependency, the build was fine.
```
[INFO] +- org.jenkins-ci.plugins:p4:jar:1.3.8:compile (optional)
[INFO] |  +- org.jenkins-ci.plugins:credentials:jar:2.3.13:compile (optional)
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -230,6 +230,11 @@
     </dependency>
 
     <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>credentials</artifactId>
+    </dependency>
+
+    <dependency>
       <artifactId>workflow-step-api</artifactId>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <optional>true</optional>
--- a/pom.xml
+++ b/pom.xml
@@ -238,6 +238,12 @@
       <artifactId>workflow-step-api</artifactId>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <optional>true</optional>
+      <exclusions>
+        <exclusion>
+          <groupId>org.jenkins-ci.plugins</groupId>
+          <artifactId>*</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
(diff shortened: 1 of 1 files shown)


warden:
@olamy I'm wondering if it is somehow possible to test this via `@JenkinsRule`?
As it seems this is not detected via the current tests, I'm assuming that it installs "the selected plugins", among which was the `Credentials plugin`. I looked at `JenkinsRule` but I didn't find the place where it is taking the `default` plugin list from.

warden:
Not final let, do not merge please.

olamy (approved):

TobiX:
@warden I added an additional test suite which runs without the optional plugins to my plugin: https://github.com/jenkinsci/dashboard-view-plugin/blob/master/pom.xml#L154-L173 - Maybe a similar setup would help catching errors like this in the future.

warden:
@TobiX that's a nice approach but if I understand correct your change, that actually does exactly the opposite. When I just checked what happens on `dashboard-view` on the test run - it runs the tests twice: once full suite + once w/o Matrix dependency.
What I wanted is the opposite, to get a compile error straight away in IDE when API  from optional dependency is used but the dependency is not directly declared. Here an api from `credentials-plugin` was used, but it was not declared a dependency as it should be, but the compilation was fine because it was transitively provided via one of the optional plugins (p4 IIRC).

TobiX:
Actually, I'd argue it would still help: Testing without optional dependencies obviously won't get an error on compile time, but the test suite would fail if you depended on functionality from a plugin you rely on transitively via an optional dependency (now that I think about it, that would probably make a good enhancement to surefire: run a site with all optional dependencies excluded).

It's a shame Maven doesn't support something like:
```
<exclude>
  <packaging>hpi</packaging>
</exclude>
```
😄

warden:
Sure, I agree, to get the tests failure that will help definitely, but with the disadvantage that the tests will run twice. My goal was to fail early and not get to the stage when someone merges in something that should evidently fail on compilation stage :-)

olamy:
this Maven plugin can detect such error (i.e used and undeclared and unused and declared) https://maven.apache.org/plugins/maven-dependency-plugin/analyze-mojo.html

warden:
> this Maven plugin can detect such error (i.e used and undeclared and unused and declared) https://maven.apache.org/plugins/maven-dependency-plugin/analyze-mojo.html

Indeed and it can even [fail the build in such case](https://maven.apache.org/plugins/maven-dependency-plugin/examples/failing-the-build-on-dependency-analysis-warnings.html).

PR #293, opened 2020-10-24
Add PCT GH action

warden:
This PR adds possibility to start a manual PCT workflow. 
@olamy you mentioned you're running PCT tests before release, let me know if this needs any more polishing.

### DIFF
--- a/.github/workflows/pct-tests.yml
+++ b/.github/workflows/pct-tests.yml
@@ -1,0 +1,32 @@
+name: PCT
+
+on:
+  workflow_dispatch:
+    inputs:
+      jenkins_ver:
+        description: 'Jenkins version'
+        required: true
+        default: '2.249.2'
+      branch:
+        description: 'Branch to test'
+        required: true
+        default: 'master'
+
+jobs:
+  pct:
+    runs-on: ubuntu-latest
+    container:
+      image: jenkins/pct
+      env:
+        CHECKOUT_SRC: "https://github.com/${{ github.repository }}.git"
+        VERSION: ${{ github.event.inputs.branch }}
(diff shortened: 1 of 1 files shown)


olamy:
how long is the run? if not too long maybe we can run it per default?

warden:
It seems I stumbled on a bug in GH workflows

warden:
It seems I stumbled on a bug in [GH workflows](https://github.community/t/workflow-dispatch-workflow-not-showing-in-actions-tab/130088/3), that's why this PR got closed, but it doesn't matter, I'll commit it later once you confirm it looks fine.

Sure, including it on by default would be nice. Do you mean though `by default` on PRs or on `master` ?
The question then is also about the version against which we want to trigger, as now it is a manual parameter.

The run takes ~10m - take a look at [sample run on my fork](https://github.com/warden/jira-plugin/actions/runs/327218400). Does the [report](https://github.com/warden/jira-plugin/actions/runs/327218400) look fine for you? If yes, I think we're good to go.

olamy:
on PRs and master.
jenkins version should be the same as the pom https://github.com/warden/jira-plugin/blob/cad40188c9557ea78b9a81a6925c64d5fb1aac00/pom.xml#L26
but not sure how to extract it. So make it easy just a comment in the pom saying if changing the version change it in pct gh action as well.

warden:
@olamy [it's alive](https://github.com/jenkinsci/jira-plugin/runs/1308529736?check_suite_focus=true), with [dynamically fetched jenkins.version](https://github.com/jenkinsci/jira-plugin/commit/b2b70d5bc40d6449ba6f2ca028d7ca0150b15c9a).

PR #299, opened 2020-11-06
[JENKINS-64148] Restore binary compatibility in JiraBuildAction

gmshake:
Restore binary compatability. See https://github.com/jenkinsci/jira-plugin/pull/283/ , issue link https://issues.jenkins-ci.org/browse/JENKINS-64148

CC @warden

### DIFF
--- a/src/main/java/hudson/plugins/jira/JiraBuildAction.java
+++ b/src/main/java/hudson/plugins/jira/JiraBuildAction.java
@@ -1,9 +1,7 @@
 package hudson.plugins.jira;
 
-import hudson.model.Action;
 import hudson.model.Run;
 import hudson.plugins.jira.model.JiraIssue;
-import java.io.Serializable;
 import java.net.URL;
 import java.util.HashSet;
 import java.util.Set;
@@ -25,6 +23,19 @@ public class JiraBuildAction implements RunAction2 {
 
   public JiraBuildAction(@Nonnull Set<JiraIssue> issues) {
     this.issues = new HashSet<>(issues);
+  }
+
+  // Leave it in place for binary compatibility.
+  /**
+   * @deprecated use {@link #JiraBuildAction(java.util.Set)} instead
+   *
+   * @param owner the owner of this action
+   * @param issues the Jira issues
+   */
+  @Deprecated
+  public JiraBuildAction(Run<?, ?> owner, @Nonnull Set<JiraIssue> issues) {
+    this(issues);
+    // the owner will be set by #onAttached(hudson.model.Run)
   }
 
   @Override


src/main/java/hudson/plugins/jira/JiraBuildAction.java
```diff
@@ -27,6 +27,11 @@ public JiraBuildAction(@Nonnull Set<JiraIssue> issues) {
     this.issues = new HashSet<>(issues);
   }
 
+  public JiraBuildAction(@Nonnull Run<?, ?> owner, @Nonnull Set<JiraIssue> issues) {
```
warden, line 30: Please mark it `@Deprecated`

src/main/java/hudson/plugins/jira/JiraBuildAction.java
```diff
@@ -27,6 +27,11 @@ public JiraBuildAction(@Nonnull Set<JiraIssue> issues) {
...
+    this.owner = owner;
```
warden, line 31: IIUC what @jglick said about `RunAction2` (quite unfriendly name btw 😆 ), since this is now set by `onLoad`, this line should be removed?

src/main/java/hudson/plugins/jira/JiraBuildAction.java
```diff
@@ -27,6 +27,11 @@ public JiraBuildAction(@Nonnull Set<JiraIssue> issues) {
...
+    this.issues = new HashSet<>(issues);
```
warden, line 32: so finally this should actually call `JiraBuildAction(issues)`

warden (changes requested):

src/main/java/hudson/plugins/jira/JiraBuildAction.java
gmshake: Checked the implementation of `Run#addAction(hudson.model.Action)`, I can confirm the `owner` is set by `jenkins.model.RunAction2#onAttached(hudson.model.Run)`. I'm OK to remove it.

src/main/java/hudson/plugins/jira/JiraBuildAction.java
gmshake: Good catch :)

src/main/java/hudson/plugins/jira/JiraBuildAction.java
gmshake: Done.

src/main/java/hudson/plugins/jira/JiraBuildAction.java
jglick: Yes, so long as the owner is not used by some method called by the constructor (recall such a situation in the `junit` plugin), you can use `onAttached`.

`RunAction` was the original marker interface but people were storing a non`transient` owner and it was causing mayhem. Jenkins convention is to add a numeric suffix to interfaces that supersede older ones.

warden: Thanks for the heads up Jesse. I'm just against any `<identifier>2` naming in any code as for _just me_ it's not [transparent](https://softwareengineering.stackexchange.com/questions/138948/how-do-you-evolve-version-an-interface) and promotes bad practice in general - but that's just me :)

warden:
Thanks @gmshake!

jglick:
Looks like this deserves an emergency plugin release.

warden:
Released in `jira-3.1.3`.

PR #162, opened 2018-08-22
expand environment variables for Issue Field Updater

romanblachman:
Solution for JENKINS-43980: Issue Field Updater unable to assign Jenkins env variable in field value for custom field id

Also, added support for multi-value field update as part of the Issue Field Updater build step.

PR #319, opened 2021-02-09
JENKINS-43980: add support for expanding variables in IssueFieldUpdaterStep

warden:
- replaces #162
- bumped Jenkins version to 2.249.3 for EnvVars to be available in workflow steps
- added a test for variable expansion


warden:
Looking good, I'll just need to do a manual test and we'll be good to go.

pom.xml
```diff
@@ -23,7 +23,7 @@
     <jira-rest-client.version>5.2.2</jira-rest-client.version>
     <fugue.version>3.0.0</fugue.version>
     <!-- jenkins -->
-    <jenkins.version>2.222.4</jenkins.version>
+    <jenkins.version>2.249.3</jenkins.version>
```
olamy, line 26: This should be part of a separate PR to have his own line in the release notes

olamy (changes requested):

pom.xml
warden: I did it in a separate commit, I think that's enough, for the release notes, we can just edit it manually and link the commit id.

olamy: We have tool to avoid manual errors/forgetting btw someone else can cut the release and forget.
but that's fine as long as you do not forget it that will be fine :)

olamy: BTW this branch has 6 commits. will you squash 5 and 1 (with jenkins version upgrade)?

warden: The previous commits need to stay (as they were picked up from the previous PR that will be auto-closed when this PR gets merged). 
Also as I understand, the merge commits cannot be squashed, they can be rebased though but we had a discussion in the past about it with jglick that's irrelevant as they shouldn't be shown after merge.. but I might be wrong here :)

olamy: Ideally we should have 1 commit (PR squashed) per feature. 
So here we should have 2 commits:
- the new feature (co-authored by you and the other contributor)
- jenkins core version required upgrade

perso I prefer to have clean PR so I use rebase only to not have those noisy merge commits.

olamy: at least update the title of the PR to have some details on core upgrade

warden: Done the squashing, looks fine to me now. 
I'll mention the core upgrade in a separate bullet point in the release notes referring to the commit id.

olamy (approved):
awesome job thanks!

PR #324, opened 2021-03-02
JENKINS-64988: Reduce log level of annotation messages

omehegan:
On a busy controller, the 'Annotating Jira id ...' messages can log thousands of entries per day, which makes it difficult to debug other issues using the service logs. I propose to reduce the log level to FINE so these messages are not logged by default.

https://issues.jenkins.io/browse/JENKINS-64988

### DIFF
--- a/src/main/java/hudson/plugins/jira/JiraChangeLogAnnotator.java
+++ b/src/main/java/hudson/plugins/jira/JiraChangeLogAnnotator.java
@@ -77,7 +77,7 @@ public class JiraChangeLogAnnotator extends ChangeLogAnnotator {
                     continue;
                 }
 
-                LOGGER.log(Level.INFO, "Annotating Jira id: ''{0}''", id);
+                LOGGER.log(Level.FINE, "Annotating Jira id: ''{0}''", id);
 
                 URL url, alternativeUrl;
                 try {


artkoshelev:
I had no strong feelings about this change but can imaging that could break someone's workflow. If only we can make logging fine-tuned in runtime.

warden:
@artkoshelev there can be [specific loggers](https://www.jenkins.io/doc/book/system-administration/viewing-logs/) that are fine-tuned to end-user's needs

warden (dismissed):

warden:
@omehegan can you explain what is your real issue with this logger? there is a lot of other logging in a busy environment I imagine, for me after re-thinking I think it should stay on INFO for visibility and I don't think that one line in the big log contributes much here.

omehegan:
@warden I'll answer your question with a question: what is the value in the logging as currently configured? Who needs to see this information in logs, and why? It's just the plugin telling you that it did something as part of its normal course of operation, but you can already tell that it did it by looking at the UI. 

In my example case, in the course of a day on this system, there were 10,765 log entries generated, and 5,344 were "Annotating Jira id."

olamy (approved):
Agree on merging this. As this seems to be a plugin developer sort of debug.

PR #343, opened 2021-05-09
Remove usages of `com.google.common.base.Objects#firstNonNull`

basil:
`com.google.common.base.Objects#firstNonNull` has been removed in recent versions of Guava, so consuming this API is a liability. This PR replaces usages of `com.google.common.base.Objects#firstNonNull` with native Java functionality to allow this plugin to be used seamlessly with any version of Guava in the future.


artkoshelev (changes requested):
This is not an equivalent change, let's have a look into `firstNonNull` desciption:
> Returns the first of two given parameters that is not null, if either is, or otherwise throws a NullPointerException.

You didn't implement the last part when both objects are `null`.

PR #368, opened 2021-08-23
get rid of guava usage, replace guava cache with caffeine, keep guava dependency here as it is needed for some tests

olamy:
Signed-off-by: Olivier Lamy <olamy@apache.org>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -213,6 +213,12 @@
       </exclusions>
     </dependency>
 
+    <dependency>
+      <groupId>io.jenkins.plugins</groupId>
+      <artifactId>caffeine-api</artifactId>
+      <version>2.9.1-23.v51c4e2c879c8</version>
+    </dependency>
+
     <!-- Jenkins plugin dependencies -->
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
--- a/src/main/java/com/atlassian/httpclient/apache/httpcomponents/SettableFuturePromiseHttpPromiseAsyncClient.java
+++ b/src/main/java/com/atlassian/httpclient/apache/httpcomponents/SettableFuturePromiseHttpPromiseAsyncClient.java
@@ -3,8 +3,6 @@ package com.atlassian.httpclient.apache.httpcomponents;
 import com.atlassian.sal.api.executor.ThreadLocalContextManager;
 import io.atlassian.util.concurrent.Promise;
 import io.atlassian.util.concurrent.Promises;
-import com.google.common.annotations.VisibleForTesting;
-import com.google.common.util.concurrent.SettableFuture;
 import org.apache.http.HttpResponse;
 import org.apache.http.client.methods.HttpUriRequest;
 import org.apache.http.concurrent.FutureCallback;
@@ -14,11 +12,11 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.util.Objects;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeoutException;
 
-import static com.google.common.base.Preconditions.checkNotNull;
 
 final class SettableFuturePromiseHttpPromiseAsyncClient<C> implements PromiseHttpAsyncClient
 {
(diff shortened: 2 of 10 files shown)


src/main/java/hudson/plugins/jira/JiraSite.java
```diff
@@ -811,7 +809,7 @@ public String getPropertyValue(final String s) {
      */
     @Nullable
     public URL getUrl() {
-        return Objects.firstNonNull(this.url, this.alternativeUrl);
+        return this.url != null ? this.url : this.alternativeUrl;
```
warden, line 812: We had multiple misinterpretations of the usage of main url and alternativeUrl, when touching this I'm wondering if we couldn't already just use `this.url` here and treat alternativeUrl as a totally separate thing used for generated links/views only (not for API communication). In that case, this method should be actually `@NotNull﻿`, if not, then I guess `CheckForNull` instead of `Nullable` per [this discussion](https://intellij-support.jetbrains.com/hc/en-us/community/posts/115000214944-Why-javax-annotation-CheckForNull-is-not-included-in-nullable-annotations-set-by-default-)

src/main/java/hudson/plugins/jira/JiraSite.java
```diff
@@ -895,22 +893,17 @@ public Pattern getIssuePattern() {
...
+            if (this.jiraSession == null) {
+                return Optional.empty();
             }
+            return Optional.ofNullable(this.jiraSession.getIssue(id));
+        });
 
-            return new JiraIssue(issue.get());
-        } catch (ExecutionException e) {
-            throw new IOException(e);
+        if (!issue.isPresent()) {
```
warden, line 903: Possible NPE?

pom.xml
```diff
@@ -213,6 +213,12 @@
       </exclusions>
     </dependency>
 
+    <dependency>
```
warden, line 216: Looking at the title of the PR, shouldn't `<groupId>com.google.guava</groupId>` be removed then? :-)

olamy: you should read the comment here ;)  https://github.com/jenkinsci/jira-plugin/blob/master/pom.xml#L142

src/main/java/com/atlassian/httpclient/apache/httpcomponents/SettableFuturePromiseHttpPromiseAsyncClient.java
```diff
@@ -30,30 +28,29 @@
...
     @Override
     public Promise<HttpResponse> execute(HttpUriRequest request, HttpContext context)
     {
     	// TODO after migrating from atlassian-util-concurrent 3.0.0 to 4.0.0 the SettableFuture.create() maybe obsolete ?
-        final SettableFuture<HttpResponse> future = SettableFuture.create();
         Future<org.apache.http.HttpResponse> clientFuture = client.execute(request, context, new ThreadLocalContextAwareFutureCallback<C>(threadLocalContextManager)
         {
+
             @Override
             void doCompleted(final HttpResponse httpResponse)
```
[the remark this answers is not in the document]
olamy: fixed with using CompletableFuture

src/main/java/hudson/plugins/jira/JiraSite.java
olamy: sounds like too much change? :)

warden: Sure, I agree, just wanted to bring this to our discussion :)

pom.xml
```diff
@@ -143,7 +143,7 @@
     	<!-- Remove this dependency and the exclusion above when jenkins provides a up-to-date guava version (currently 11.0.1) -->
```
warden, line 143: Does this comment still make sense?

warden (approved):

pom.xml
```diff
@@ -139,11 +139,10 @@
...
     </dependency>
     <dependency>
-    	<!-- version 18 is the first one containing com.google.common.base.MoreObjects.ToStringHelper which is used by com.atlassian.jira.rest.client.api.domain.Issue -->
     	<!-- Remove this dependency and the exclusion above when jenkins provides a up-to-date guava version (currently 11.0.1) -->
-        <groupId>com.google.guava</groupId>
-        <artifactId>guava</artifactId>
-        <version>18.0</version>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+      <version>30.1.1-jre</version>
```
MRamonLeon, line 145: is it used in producation or just tests? @olamy Wondering to add test scope

PR #401, opened 2022-01-11
[JENKINS-64083] Updated JiraSite to use global credentials when no job-specific credentials are defined

kevin-wimmer:
This change includes the update from Carsten Pfeiffer to address JENKINS-64083 as well as new test method in JiraSiteTest to verify the update.


ashkulz:
Thanks, @kevin-wimmer! I can confirm that this fixes [JENKINS-62706](https://issues.jenkins.io/browse/JENKINS-62706) as well :tada:

PR #404, opened 2022-01-21
remove guava as provided by core

olamy:
Guava is not up2date in Jenkins LTS so all the hard-coded hacks can be removed.

PR #408, opened 2022-02-03
Bump core baseline past Guava upgrade

basil:
Alternative to #404.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>4.31</version>
+    <version>4.33</version>
     <relativePath />
   </parent>
 
@@ -21,9 +21,9 @@
     <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
     <java.level>8</java.level>
     <jira-rest-client.version>5.2.2</jira-rest-client.version>
-    <fugue.version>3.0.0</fugue.version>
+    <fugue.version>4.7.2</fugue.version>
     <!-- jenkins -->
-    <jenkins.version>2.277.4</jenkins.version>
+    <jenkins.version>2.324</jenkins.version>
     <!-- security spotbugs -->
     <spotbugs.failOnError>false</spotbugs.failOnError>
   </properties>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -246,13 +264,6 @@
     <dependency>
       <artifactId>workflow-step-api</artifactId>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
-      <optional>true</optional>
```
rantoniuk, line 249: Why is optional removed?

pom.xml
```diff
@@ -324,8 +331,8 @@
     <dependencies>
       <dependency>
         <groupId>io.jenkins.tools.bom</groupId>
-        <artifactId>bom-2.277.x</artifactId>
-        <version>984.vb5eaac999a7e</version>
+        <artifactId>bom-weekly</artifactId>
```
rantoniuk, line 334: I think we prefer depending on a specific line of bom, rather than weekly, unless this is some 'global' approach change that I'm not aware of?

rantoniuk (commented):
I like this! 
Some minor questions and let's wait for @olamy to review.

pom.xml
basil: There's already a non-optional dependency on `workflow-job`, which itself has a non-optional dependency on `workflow-step-api`. So in practice, the dependency on `workflow-step-api` was already non-optional. Removing `optional` makes this more clear.

pom.xml
basil: But there is no LTS release available yet with the Guava upgrade, so there is no BOM line available yet with the Guava upgrade. The only Jenkins release with the Guava upgrade is a weekly, and the correct BOM `artifactId` for pre-LTS weekly releases is `bom-weekly`. I suggest you read the documentation in `jenkinsci/bom`.

basil:
> let's wait for @olamy to review.

Why do we need to wait?

pom.xml
```diff
@@ -21,9 +21,9 @@
     <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
     <java.level>8</java.level>
     <jira-rest-client.version>5.2.2</jira-rest-client.version>
-    <fugue.version>3.0.0</fugue.version>
+    <fugue.version>4.7.2</fugue.version>
     <!-- jenkins -->
-    <jenkins.version>2.277.4</jenkins.version>
+    <jenkins.version>2.324</jenkins.version>
```
olamy, line 26: I definitely prefer wait for next LTS including guava upgrade. it's coming soon.
the plugin is currently working with guava upgrade.
I don't why we should rush the upgrade and prevent a lot users using the new releases of the plugin.

olamy (changes requested):

olamy:
> > let's wait for @olamy to review.
> 
> Why do we need to wait?

see my answer here https://github.com/jenkinsci/jira-plugin/pull/408#discussion_r799905054
there is definitely no rush for such core version upgrade as this will prevent a lot users to upgrade...
And this change as it change a lot of dependencies used by atlassian jira client so it need to be tested (manually as we don't have yet some automatic testing) on some real jira instance.

olamy:
@basil No need to close it but turn it into a draft so we can merge it when we have the LTS ready

PR #413, opened 2022-02-15
avoid NPE when outsite of a job context

olamy:
Signed-off-by: Olivier Lamy <olamy@apache.org>

### DIFF
--- a/src/main/java/hudson/plugins/jira/JiraMailAddressResolver.java
+++ b/src/main/java/hudson/plugins/jira/JiraMailAddressResolver.java
@@ -4,8 +4,8 @@ import hudson.Extension;
 import hudson.model.Job;
 import hudson.model.User;
 import hudson.tasks.MailAddressResolver;
-import jenkins.model.Jenkins;
 import org.kohsuke.stapler.Stapler;
+import org.kohsuke.stapler.StaplerRequest;
 
 import java.util.List;
 import java.util.logging.Logger;
@@ -35,11 +35,10 @@ public class JiraMailAddressResolver extends MailAddressResolver {
         String username = u.getId();
 
         Job<?, ?> job = null;
-        try {
-            job = Stapler.getCurrentRequest().findAncestorObject(Job.class);
-        } catch (NullPointerException e) {
-            // sadly but outside of a context job this throw an exception
-            LOGGER.fine("NPE trying to find the Job of the current request");
+
+        StaplerRequest req = Stapler.getCurrentRequest();
+        if(req != null) {
+            job = req.findAncestorObject(Job.class);
         }
 
         List<JiraSite> sites = job == null ? JiraGlobalConfiguration.get().getSites() : JiraSite.getJiraSites(job);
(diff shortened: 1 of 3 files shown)


src/main/java/hudson/plugins/jira/JiraMailAddressResolver.java
```diff
@@ -32,9 +34,17 @@ public String findMailAddressFor(User u) {
         }
         String username = u.getId();
 
-        Job<?, ?> job = Stapler.getCurrentRequest().findAncestorObject( Job.class);
+        Job<?, ?> job = null;
+        try {
+            job = Stapler.getCurrentRequest().findAncestorObject(Job.class);
+        } catch (NullPointerException e) {
+            // sadly but outside of a context job this throw an exception
```
rantoniuk, line 41: Since there is not test added... what is the usecase (i.e. what are the circumstances under which this error is thrown?)

olamy: I can try to add some tests but not sure how. there is no much test currently for this part.
Just create your own SecurityListener extension (https://javadoc.jenkins.io/jenkins/security/SecurityListener.html)
just after login on the home page the method authenticated2 is fired and you're not part of any job context.

this make me think we could add some caching for the getUser method from Jira? nothing really urgent but why not.

olamy: > I understand that you're able to reproduce this error, so... that's the test (or scenario) I was asking about :-) There is already a test for the [resolver](https://github.com/jenkinsci/jira-plugin/blob/master/src/test/java/hudson/plugins/jira/MailResolverTest.java) that is exactly checking the null in case it was disabled via properties. This NPE is different though, so I'm trying to understand why this resolver is fired when you only log in..? I'm happy with the fix itself though.

Just create your own SecurityListener extension (https://javadoc.jenkins.io/jenkins/security/SecurityListener.html) and in try to figure out the logged in user email with such code 
```
            User user = User.getById(userId, false);
            if (user != null) {
                Mailer.UserProperty emailProperty = user.getProperty(Mailer.UserProperty.class);
                if (emailProperty != null && emailProperty.getAddress() != null) {
                    email = emailProperty.getAddress();
                }
            }
```
trying to get the user email will fire this NPE. 
For some reason in this case (I haven't really dig into it but I presume it's because we are in the context of spring security handlers and not yet stapler) 
```
StaplerRequest req = Stapler.getCurrentRequest();
here req is null 
```

> 
> Adding caching might be good as a separate feature, not as a fix to this scenario, because it wouldn't fix it, only make it harder to debug.

yes definitely it was just loud thinking :) and as I said nothing urgent

olamy: @rantoniuk  test added

olamy:
@rantoniuk thanks for merging, I will have a look at other PRs and release really soon :)

PR #415, opened 2022-02-19
Update Jenkinsfile to runPct

olamy:


rantoniuk:
@olamy does adding PCT tests for this plugin still make sense?

rantoniuk:
I don't see `PCT` in https://github.com/jenkinsci/jenkins/blob/master/Jenkinsfile so I guess this is no longer neeeded or needs refactoring.

Issue #389, opened 2021-11-07
Consider removing PowerMock


rantoniuk:
PowerMock was removed from [plugin-pom](https://github.com/jenkinsci/plugin-pom/pull/442#issue-1002043253) due to supposed lack of activity (see #388). 
Consider pros and cons of removing it from current tests.

artkoshelev:
I'm all in to replace existing usages with Mockito and use 1 mocking framework instead of 2

rantoniuk:
@artkoshelev would you want to devote some time on this? :-)

artkoshelev:
maybe =)

PR #462, opened 2022-07-31
Remove power-mock

artkoshelev:
### What:
Remove powermock, slightly refactor `JiraSite` to make tests work with Mockito

### Why:
* For consistency, to have a single mocking framework
* A requirement to modify object state via reflection (which Powermock is doing) to test it, indicates wrong design and should be fixed

### Links
https://github.com/jenkinsci/jira-plugin/issues/389

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -319,12 +319,6 @@
       <scope>test</scope>
     </dependency>
     <dependency>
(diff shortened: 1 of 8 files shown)


rantoniuk:
Thanks a lot! 🙏

Issue #460, opened 2022-07-25
Be able to use build/environment parameters in jql selector


mrvcmrvc:
### Feature Request

Plugin Version => 3.7

When using **Jira: Update relevant issues** or **Jira: Issue custom field updater**, they are not able to read build parameters in **JQL selector** and in **explicit selector**.

What I am trying to do is, instead if writing **key = "EXAMPLE-1"**, I want to be able to write **key = ${IssueId}** which IssueId is a parameter defined in the job.

- Since $ is preserved for Jira, I tried to use escape character as well (**key = \\${IssueId} or key = \\\\${IssueId}**)
- I thought maybe the issue is related with missing quotation marks so I tried **key = "${IssueId}"** or **key = "\\${IssueId}"**
- I tried to add quotation marks in the build parameter (IssueId) instead of jql query (eg. if the issue id is TEST-123, then the parameter on the job is "TEST-123")

None of them worked. If I wrote a hardcoded issue id instead of a build parameter, it works flawlessly.

---

I noticed a similar feature is requested for **IssueFieldUpdateStep** to be able to use build/environment variables in field value when updating a custom field in JIRA with a given value. It has been implemented with the release of [3.2](https://github.com/jenkinsci/jira-plugin/releases/tag/jira-3.2), and here is the [commit message](https://github.com/jenkinsci/jira-plugin/pull/319/commits/12dd7e81cb2e05886b4a5bc96aacf5eb345df4e4#diff-022e48912309f7bc4e6ed655687f0cc1ee3754508cc48cdb646346018dc9a6f7) that contains the change.

---

I also checked the test code for **JqlIssueSelector**;

- It creates a jql selector with a jql containing hardcoded issue id,
![Screenshot 2022-07-25 at 21 48 56](https://user-images.githubusercontent.com/11190664/180871674-de58f463-7cba-4af3-8539-8bf7a9434da9.png)
- Selector sends the jql to session in findIssueIds() method,
![Screenshot 2022-07-25 at 21 51 14](https://user-images.githubusercontent.com/11190664/180871854-1a9ff683-90d0-49ad-ad34-d4976d250ad7.png)
- Session passes it to JiraRestService,
- JiraRestService searches issues by passing jql to JiraRestClient
![Screenshot 2022-07-25 at 21 52 25](https://user-images.githubusercontent.com/11190664/180872015-bd7e2bf5-862b-4402-bb46-9a8af4fe3415.png)

[...]

PR #469, opened 2022-08-01
Environment Expansion for issue Ids on various steps and selectors

mrvcmrvc:
- Extract environment expansion to its own class (EnvironmentExpander)
- Call EnvironmentExpander from various steps and selectors
    - IssueFieldUpdateStep
    - ExplicitIssueSelector
    - JqlIssueSelector

The "Environment variable expanding logic was already implemented". [Here](https://github.com/jenkinsci/jira-plugin/issues/460) is the detailed explanation of the issue.

What I did is, I moved the logic to its own class and used it in JqlIssueSelecter and ExplicitIssueSelector to parse it with environment variables before sending the data to jira service.

The logic was previously implemented in IssueFieldUpdateStep, so my changes on that class was just refactoring to use the new class.

---

> No relevant pull requests etc.

### DIFF
--- a/src/main/java/hudson/plugins/jira/pipeline/IssueFieldUpdateStep.java
+++ b/src/main/java/hudson/plugins/jira/pipeline/IssueFieldUpdateStep.java
@@ -10,6 +10,7 @@ import hudson.model.AbstractProject;
 import hudson.model.Result;
 import hudson.model.Run;
 import hudson.model.TaskListener;
+import hudson.plugins.jira.EnvironmentExpander;
 import hudson.plugins.jira.JiraSession;
 import hudson.plugins.jira.JiraSite;
 import hudson.plugins.jira.Messages;
@@ -20,7 +21,6 @@ import hudson.tasks.Builder;
 import hudson.util.FormValidation;
 import java.io.IOException;
 import java.io.PrintStream;
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Set;
@@ -121,16 +121,13 @@ public class IssueFieldUpdateStep extends Builder implements SimpleBuildStep {
         List<JiraIssueField> fields = Collections.singletonList(
             new JiraIssueField(
                 prepareFieldId(getFieldId()),
-                getFieldValue()
+                EnvironmentExpander.expandVariable(getFieldValue(), env)
             )
         );
-
-        fields = expandVariables(fields, env);
 
         for (String issue : issues) {
             submitFields(session, issue, fields, logger);
         }
-
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


mrvcmrvc:
I have no idea how to fix SonarCloud check, help would be very appreciated 🙌

src/main/java/hudson/plugins/jira/pipeline/IssueFieldUpdateStep.java
```diff
@@ -20,7 +21,6 @@
 import hudson.util.FormValidation;
 import java.io.IOException;
 import java.io.PrintStream;
-import java.util.ArrayList;
```
mrvcmrvc, line 23: It was not being used, so I removed it

src/main/java/hudson/plugins/jira/pipeline/IssueFieldUpdateStep.java
```diff
@@ -140,19 +137,6 @@ public void perform(Run<?, ?> run, FilePath workspace, Launcher launcher, TaskLi
         this.perform(run, workspace, run.getEnvironment(listener), launcher, listener);
     }
 
-    private List<JiraIssueField> expandVariables(List<JiraIssueField> fields, EnvVars envVars) {
```
mrvcmrvc, line 143: This was interesting, I checked it and fields parameter is always a list with 1 element in it. The logic does not support having multiple fields.

Because, we are asking a field id from user and do some string operation on it such as adding "customField_" prefix etc and creating a JiraIssueField with it and add it to this list.

If user enters multiple field ids, the system wont be able to make any string operation on it and it will fail. So this step always works with 1 field id, so I am not sure why we are creating a list and sending it to expand variables, so I got rid of it as well. Let me know if you think we need it back

mrvcmrvc: I checked the test code but we are not sending multiple custom fields to the step. We are sending them one by one and parsing it one by one.

```
    public String prepareFieldId(String fieldId) {
        String prepared = fieldId;
        if (!prepared.startsWith("customfield_"))
            prepared = "customfield_" + prepared;
        return prepared;
    }
```

This is how customfield is parsed in the step, it accepts only String and not String[], so on the test code we are sending the list of String one by one and comparing if it is correct or not each time. In this case it is correct. On the Jenkins interface, we write "customfield_XXX" or "XXX" and step correctly parses it to "customfield_XXX".

So, there is actually no way to send multiple customfields on the step. Step works with only 1 customfield. That is why test passed, because we are not sending multiple fields at the same time.

mrvcmrvc: Also another thing I noticed is, during the test we create 2 lists. 1 before the test and 1 after the test. And what we do is sending each element in the "before test" one by one to the step and comparing the outcome with the "after test" list. Here are the lists;

```
    List<String> field_test = Arrays.asList("10100", "customfield_10100", "field_10100");

    List<String> field_after =
        Arrays.asList("customfield_10100", "customfield_10100", "customfield_field_10100");
```

Since test is passing, that means the last item "field_10100" is returning as "customfield_field_10100", but I thought custom fields on Jira are always numerical values. If that is the case, there is no way a field id can be "customfield_field_10100". Altough I am not so sure, so please correct me if I am wrong.

mrvcmrvc:
@rantoniuk Can you also guide me what the issue is about sonarCloud check? I am not sure what it is complaining

rantoniuk:
@mrvcmrvc don't worry about SonarCloud, I'll handle that.
Just add the tests and it will re ready to go :)

PR #471, opened 2022-08-02
Remove foreign `atlassian-public` repository

jglick:
Jenkins plugins and other components should refer only to the Jenkins Artifactory (and implicitly anything it mirrors including of course Central), never to other Maven repositories. https://repo.jenkins-ci.org/ui/repos/tree/General/public aggregates https://repo.jenkins-ci.org/ui/repos/tree/General/atlassian which mirrors https://maven.atlassian.com/content/repositories/atlassian-public/
 so there should be no need to explicitly specify it. Amends 0834935028b987bdb19bf9d672a684111b278526. Trying to fix a problem running https://github.com/jenkinsci/maven-hpi-plugin/pull/336 depending on mirror settings, though it does not seem to work anyway.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -374,19 +374,6 @@
         <enabled>false</enabled>
       </snapshots>
     </repository>
-    <repository>
-      <id>atlassian-public</id>
-      <url>https://packages.atlassian.com/mvn/maven-public/</url>
-      <releases>
-        <enabled>true</enabled>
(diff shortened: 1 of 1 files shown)


rantoniuk (approved):

jglick:
Update: it seems that Artifactory mirror is not actually used here; at least https://search.maven.org/artifact/com.atlassian.pom/base-pom/6.1.2/pom is in Central.

PR #477, opened 2022-08-23
fix flaky CliParameterTest on Windows

jtnord:
the build was started but it did not wait for the build to complete.
On windows this would cause a failure to cleanup in the teardown of the
Jenkins instance as the build log file would still be in use (sometimes)

Fixes the issue by [adding the -s flag](https://www.jenkins.io/doc/book/managing/cli/) to the build command so the cli
command will not return until the build has completed.

fixes for example https://ci.jenkins.io/job/Plugins/job/jira-plugin/job/master/354/testReport/junit/hudson.plugins.jira/CliParameterTest/windows_11___Build__windows_11____jiraVersionParameterViaCli/


rantoniuk:
Thanks for this fix!
I tried to find the issue with those flaky tests multiple times but nobody was able to point me to the any sensible reason for it.

PR #478, opened 2022-08-23
remove sonar analysis because it's always failing as false negative

olamy:
this is always failing and marking PR checks as failed whereas it's not.
As nobody reads the reports neither `fix` what is reported, maybe we can consider this as noise.


rantoniuk:
This is always failing as I have just introduced it and I need to fix an issue with the unsupported GH action workflow.
I'll update the workflow action to not fail the build though.

olamy:
> This is always failing as I have just introduced it and I need to fix an issue with the unsupported GH action workflow. I'll update the workflow action to not fail the build though.

why have you introduced this without a PR? 
at least to ask POV from other contributing here?
Perso I feel it's like a new toy trying but nobody will read the report and fix the "issues" which can be a lot false positive....

olamy:
@rantoniuk I will merge this to fix the checks.
PLEASE PROVIDE SUCH VIA A PR!!
because you pushed something broken directly to master!!! and leave it as is broken for weeks!!

Issue #497, opened 2022-12-09
Option to use a Authorization: Bearer instead of Basic


hakamairi:
### What feature do you want to see added?

Hi,
It would be great to be able to use Personal Access Token for all the communication with Jira.
On the self hosted instance I use there's no Basic Auth allowed, but there's an option to use Personal Access Token with `-H "Authorization: Bearer %token%"`
It would be great if the plugin supported it as an option from configuration page.

ianpascoe19:
I agree this would be extremely nice to have for higher security jira instances

faminepq:
I'd like to have it there as well.

EliaBracciSumo:
Hi! I implemented that 👍  If you want you can see the implementation [here](https://github.com/jenkinsci/jira-plugin/pull/521)

rantoniuk:
Released in [3.10](https://github.com/jenkinsci/jira-plugin/releases/tag/jira-3.10)

PR #521, opened 2023-03-27
Jira Bearer Authentication

EliaBracciSumo:
Hi team! 

The intent of this PR is to allow `JiraRestService` to work with Bearer Token authentication!

To do that I implemented new methods and classes:
- [Added `JiraRestService` constructor that accept bearer token authentication](https://github.com/jenkinsci/jira-plugin/pull/521/commits/8021b08aa4a274d61a29e1c31ea1932af6c56856): doing that, I also created private method `buildBaseApiPath` to avoid to repeat code and make it more readable
- [Added class BearerHttpAuthenticationHandler](https://github.com/jenkinsci/jira-plugin/pull/521/commits/ee90ba24759230f493515365313b634e67782080): this class implements `AuthenticationHandler` class and it will be used as authentication method
- [Added token in `JiraConfig` and added bearer auth tests class](https://github.com/jenkinsci/jira-plugin/pull/521/commits/74958604cc0edd4e76dab61ef18364ea54336fd1): added a new `token` variable in `JiraConfig` class and related `jira.properties`. Added test class to tests implemented code

This implementation could be a starting point for #497

### DIFF
--- a/src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
+++ b/src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
@@ -25,7 +25,7 @@
     <f:checkbox title="${%Update Relevant Jira Issues For All Build Results}" field="updateJiraIssueForAllStatus" />
   </f:entry>
   <f:entry>
-     <f:checkbox title="${%%Use Bearer authentication}" field="useBearerAuth" />
+     <f:checkbox title="${%Use Bearer authentication}" field="useBearerAuth" />
   </f:entry>
   <f:entry title="${%Credentials}" field="credentialsId">
     <c:select />
--- a/src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
+++ b/src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
@@ -9,8 +9,8 @@
   <f:invisibleEntry>
     <f:checkbox title="${%Use HTTP authentication instead of normal login}" field="useHTTPAuth" />
   </f:invisibleEntry>
-  <f:entry>
-    <f:checkbox title="${%Use Bearer authentication instead of Basic authentication}" field="useBearerAuth" description="${%site.useBearerAuth}" />
+  <f:entry description="${%site.useBearerAuth}">
+    <f:checkbox title="${%Use Bearer authentication instead of Basic authentication}" field="useBearerAuth" />
   </f:entry>
   <f:entry>
     <f:checkbox title="${%Supports Wiki notation}" field="supportsWikiStyleComment" />
--- a/src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
+++ b/src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
@@ -12,6 +12,14 @@
   <f:entry>
      <f:checkbox title="${%Use Bearer authentication}" field="useBearerAuth" />
   </f:entry>
+  <f:entry title="${‰Check the authentication method you want to use}" field="authenticationMethod">
+    <f:entry>
+      <f:radio name="useHTTPAuth" title="${%Use HTTP authentication instead of normal login}" checked="true" />
+    </f:entry>
+    <f:entry>
+      <f:radio name="useBearerAuth" title="${%Use Bearer authentication}" />
+    </f:entry>
+  </f:entry>
   <f:entry>
     <f:checkbox title="${%Supports Wiki notation}" field="supportsWikiStyleComment" />
   </f:entry>
@@ -26,9 +34,6 @@
   </f:entry>
   <f:entry>
     <f:checkbox title="${%Update Relevant Jira Issues For All Build Results}" field="updateJiraIssueForAllStatus" />
-  </f:entry>
-  <f:entry>
-     <f:checkbox title="${%Use Bearer authentication}" field="useBearerAuth" />
   </f:entry>
   <f:entry title="${%Credentials}" field="credentialsId">
     <c:select />
--- a/src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
+++ b/src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
@@ -10,7 +10,7 @@
     <f:checkbox title="${%Use HTTP authentication instead of normal login}" field="useHTTPAuth" />
   </f:invisibleEntry>
   <f:entry>
-     <f:checkbox title="${%Use Bearer authentication instead of Basic authentication}" field="useBearerAuth" />
+     <f:checkbox title="${%Use Bearer authentication instead of Basic authentication (Note: Bearer authentication is only supported in Jira Server, for Jira Cloud leave this unchecked)}" field="useBearerAuth" />
   </f:entry>
   <f:entry>
     <f:checkbox title="${%Supports Wiki notation}" field="supportsWikiStyleComment" />
--- a/src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
+++ b/src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
@@ -6,16 +6,10 @@
   <f:entry title="Link URL" field="alternativeUrl" description="${%site.alternativeUrl}">
     <f:textbox />
   </f:entry>
-  <f:entry>
-    <f:checkbox title="${%Use HTTP authentication instead of normal login}" field="useHTTPAuth" />
-  </f:entry>
-  <f:entry>
-     <f:checkbox title="${%Use Bearer authentication}" field="useBearerAuth" />
-  </f:entry>
-  <f:entry name="authenticationMethod" title="${%Check the authentication method you want to use}" field="authenticationMethod">
-    <select name="authenticationMethod">
-        <option value="useHTTPAuth" selected="${it.useHTTPAuth}">Use HTTP authentication</option>
-        <option value="useBearerAuth" selected="${it.useBearerAuth} >Use Bearer authentication</option>
+  <f:entry name="useBearerAuth" title="${%Check the authentication method you want to use}" field="useBearerAuth">
+    <select name="useBearerAuth">
+        <option value="false" selected="!${useBearerAuth}">Use HTTP authentication</option>
+        <option value="true" selected="${useBearerAuth}" >Use Bearer authentication</option>
     </select>
 </f:entry>
   <f:entry>
@@ -59,7 +53,7 @@
   </f:entry>
   <f:entry>
     <f:validateButton title="${%Validate Settings}"
-            method="validate" with="url,credentialsId,groupVisibility,roleVisibility,${authenticationMethod == 'useHTTPAuth'},alternativeUrl,timeout,readTimeout,threadExecutorNumber,${authenticationMethod == 'useBearerAuth'}" />
+            method="validate" with="url,credentialsId,groupVisibility,roleVisibility,useBearerAuth,alternativeUrl,timeout,readTimeout,threadExecutorNumber,useBearerAuth" />
   </f:entry>
   <f:entry title="">
     <div align="right">
--- a/src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
+++ b/src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
@@ -10,7 +10,7 @@
     <f:checkbox title="${%Use HTTP authentication instead of normal login}" field="useHTTPAuth" />
   </f:invisibleEntry>
   <f:entry>
-    <f:checkbox title="${%Use Bearer authentication instead of Basic authentication}" field="useBearerAuth" description="${%Note: Bearer authentication is only supported in Jira Server, for Jira Cloud leave this unchecked}" />
+    <f:checkbox title="${%Use Bearer authentication instead of Basic authentication}" field="useBearerAuth" description="${%site.useBearerAuth}" />
   </f:entry>
   <f:entry>
     <f:checkbox title="${%Supports Wiki notation}" field="supportsWikiStyleComment" />
--- a/src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
+++ b/src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
@@ -59,7 +59,7 @@
   </f:entry>
   <f:entry>
     <f:validateButton title="${%Validate Settings}"
-            method="validate" with="url,credentialsId,groupVisibility,roleVisibility,${authenticationMethod.value == 'useHTTPAuth'},alternativeUrl,timeout,readTimeout,threadExecutorNumber,${authenticationMethod.value == 'useBearerAuth'}" />
+            method="validate" with="url,credentialsId,groupVisibility,roleVisibility,${authenticationMethod == 'useHTTPAuth'},alternativeUrl,timeout,readTimeout,threadExecutorNumber,${authenticationMethod == 'useBearerAuth'}" />
   </f:entry>
   <f:entry title="">
     <div align="right">
--- a/src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
+++ b/src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
@@ -12,14 +12,6 @@
   <f:entry>
      <f:checkbox title="${%Use Bearer authentication}" field="useBearerAuth" />
   </f:entry>
-  <f:entry title="${‰Check the authentication method you want to use}" field="authenticationMethod">
-    <f:entry>
-      <f:radio name="useHTTPAuth" title="${%Use HTTP authentication instead of normal login}" checked="true" />
-    </f:entry>
-    <f:entry>
-      <f:radio name="useBearerAuth" title="${%Use Bearer authentication}" />
-    </f:entry>
-  </f:entry>
   <f:entry>
     <f:checkbox title="${%Supports Wiki notation}" field="supportsWikiStyleComment" />
   </f:entry>
--- a/src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
+++ b/src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
@@ -10,8 +10,7 @@
     <f:checkbox title="${%Use HTTP authentication instead of normal login}" field="useHTTPAuth" />
   </f:invisibleEntry>
   <f:entry>
-    <f:checkbox title="${%Use Bearer authentication instead of Basic authentication}" field="useBearerAuth"
-      description="${%(Note: Bearer authentication is only supported in Jira Server, for Jira Cloud leave this unchecked)}" />
+    <f:checkbox title="${%Use Bearer authentication instead of Basic authentication}" field="useBearerAuth" description="${%Note: Bearer authentication is only supported in Jira Server, for Jira Cloud leave this unchecked}" />
   </f:entry>
   <f:entry>
     <f:checkbox title="${%Supports Wiki notation}" field="supportsWikiStyleComment" />
--- a/src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
+++ b/src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
@@ -10,7 +10,8 @@
     <f:checkbox title="${%Use HTTP authentication instead of normal login}" field="useHTTPAuth" />
   </f:invisibleEntry>
   <f:entry>
-     <f:checkbox title="${%Use Bearer authentication instead of Basic authentication (Note: Bearer authentication is only supported in Jira Server, for Jira Cloud leave this unchecked)}" field="useBearerAuth" />
+    <f:checkbox title="${%Use Bearer authentication instead of Basic authentication}" field="useBearerAuth"
+      description="${%(Note: Bearer authentication is only supported in Jira Server, for Jira Cloud leave this unchecked)}" />
   </f:entry>
   <f:entry>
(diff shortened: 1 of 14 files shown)


EliaBracciSumo:
Hi @rantoniuk, can you please review my code? Thank you in advance!

src/test/java/JiraTesterBearerAuth.java
```diff
@@ -0,0 +1,123 @@
...
+                .createWithAuthenticationHandler(uri, handler);
+
+        final JiraRestService restService = new JiraRestService(uri, jiraRestClient, JiraConfig.getToken(), JiraSite.DEFAULT_TIMEOUT);
+
+        final String projectKey = "TESTPROJECT";
+        final String issueId = "TESTPROJECT-425";
+        final Integer actionId = 21;
+
+        final Issue issue = restService.getIssue(issueId);
+        System.out.println("issue:" + issue);
```
rantoniuk, line 40: Tests should have assertions instead of printing out to the console

src/test/java/JiraTesterBearerAuth.java
```diff
@@ -0,0 +1,123 @@
...
+//        System.out.println(" >>> update issue TESTPROJECT-425");
+//        restService.updateIssue(issueId, Collections.singletonList(releaseVersion));
+
+//        final Issue updatedIssue = restService.progressWorkflowAction(issueId, actionId);
+//        System.out.println("Updated issue:" + updatedIssue);
+
+
+
+        for(int i=0;i<10;i++){
+            callUniq( restService );
```
rantoniuk, line 100: Why is the loop needed / what is that test case trying to achieve? I would prefer this one big method to be splitted into smaller test cases with relevant assertions because otherwise it's hard to say what you're testing - see other [test cases](https://github.com/jenkinsci/jira-plugin/blob/master/src/test/java/hudson/plugins/jira/JiraSiteTest.java)

rantoniuk (changes requested):
Hi!

Thank you for contributing, really great to see this incoming PR. I made some small comments about moving files to better places, but apart from that I think this PR is not yet ready for review:

1. the test case is actually not testing anything at the moment. You don't do any assertions and notice in the PR checks menu that the code coverage dropped:
` Line: 59.00% (-0.23% against target branch)`
1. I don't see any changes to [config.jelly](https://github.com/jenkinsci/jira-plugin/blob/master/src/main/resources/hudson/plugins/jira/JiraSite/config.jelly), so I'm not sure how this new constructor is actually used and how the site should be configured via UI in this case? Maybe the related [credentialsId](https://github.com/jenkinsci/jira-plugin/blob/master/src/main/resources/hudson/plugins/jira/JiraSite/help-credentialsId.html) need to be updated as well.

Take a look at:
- [Jelly documentation](https://www.jenkins.io/doc/developer/forms/jelly-form-controls/)
- [i18n rules for forms](https://www.jenkins.io/doc/developer/internationalization/i18n-source-code/)

Hope that helps!

src/test/java/JiraTesterBearerAuth.java
EliaBracciSumo: Thank you very much for your review and your suggestions!! 🙏  What I did is replicate code inside [JiraTester](https://github.com/jenkinsci/jira-plugin/blob/master/src/test/java/JiraTester.java) class. Do you think that add another class similar to [JiraRestServiceTest](https://github.com/jenkinsci/jira-plugin/blob/master/src/test/java/hudson/plugins/jira/JiraRestServiceTest.java) is enough? 🤔  New method `buildBaseApiPath` is implicit tested by already existing test methods inside `JiraRestServiceTest` class

rantoniuk: Ah I forgot about this _manual_ test class created long time ago ;-) 
Take a look at [JiraSiteTest](https://github.com/jenkinsci/jira-plugin/blob/master/src/test/java/hudson/plugins/jira/JiraSiteTest.java), I think that after updating the UI, adapting this Test class should be enough here.

EliaBracciSumo: Do you think that token should be passed as username and password or secret text?

EliaBracciSumo:
I updated `JiraSite` class and `config.jelly` with new boolean variable `useBearerAuth`. Furthermore, I have some considerations:
- I implemented `JiraSessionFactory` class: let me say, that's not a real factory but I think that this pattern can fit the use case
- `JiraRestServiceBearerAuthTest` has been created and adapted with new `JiraRestService` constructor
- I added `useBearerAuth` in `config.jelly` and `JiraSite` builder (used by form `doValidate`). However, it seems that constructor used by `builder` pattern is an old one: do I have to update it too?

src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
```diff
@@ -24,6 +24,9 @@
   <f:entry>
     <f:checkbox title="${%Update Relevant Jira Issues For All Build Results}" field="updateJiraIssueForAllStatus" />
   </f:entry>
+  <f:entry>
+     <f:checkbox title="${%Use Bearer authentication}" field="useBearerAuth" />
```
rantoniuk, line 28: I wonder, does it make sense to have both auth options: `useHTTPAuth` and `useBearerAuth` as checkbox? What happens if someone fills in both? 
If it does, I'd suggest moving those checkboxes to be together in the UI.
If it doesn't, then you can use Radio/option?

EliaBracciSumo:
> I updated `JiraSite` class and `config.jelly` with new boolean variable `useBearerAuth`. Furthermore, I have some considerations:
> 
> * I implemented `JiraSessionFactory` class: let me say, that's not a real factory but I think that this pattern can fit the use case
> * `JiraRestServiceBearerAuthTest` has been created and adapted with new `JiraRestService` constructor
> * I added `useBearerAuth` in `config.jelly` and `JiraSite` builder (used by form `doValidate`). However, it seems that constructor used by `builder` pattern is an old one: do I have to update it too? Basically, it's not clear to me how to grant backward compatibility for `JiraSite` constructor... If I update the builder one, I have to update all deprecated constructor and adjust `JiraSiteTest`

I updated `JiraSite` builder by removing builder method for `useBearerAuth` and added call to `setUseBearerAuth` method.

src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
EliaBracciSumo: I wasn't able to find where `useHTTPAuth ` is used 🤔

rantoniuk:
> I updated `JiraSite` class and `config.jelly` with new boolean variable `useBearerAuth`. Furthermore, I have some considerations:
> 
> * I implemented `JiraSessionFactory` class: let me say, that's not a real factory but I think that this pattern can fit the use case
> * `JiraRestServiceBearerAuthTest` has been created and adapted with new `JiraRestService` constructor
> * I added `useBearerAuth` in `config.jelly` and `JiraSite` builder (used by form `doValidate`). However, it seems that constructor used by `builder` pattern is an old one: do I have to update it too? Basically, it's not clear to me how to grant backward compatibility for `JiraSite` constructor... If I update the builder one, I have to update all deprecated constructor and adjust `JiraSiteTest`

That looks great!

I would love you to update everything but that's out of scope for a single PR and would make a mess to review. I propose to do it "in the old way" now and handle the deprecations in another PR when this is merged and you still have power to contribute :)

EliaBracciSumo:
> I would love you to update everything but that's out of scope for a single PR and would make a mess to review. I propose to do it "in the old way" now and handle the deprecations in another PR when this is merged and you still have power to contribute :)

Got it 👍  In that case, how do you suggest to proceed?

rantoniuk:
When you feel your change is ready, let me know. Note that you can use the incremental build to do a test run in your sandbox instance to see if it works as expected (link is in the checks panel)

src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
EliaBracciSumo: @rantoniuk as I supposed, this variable it's no more used. It was introduced in this commit https://github.com/jenkinsci/jira-plugin/commit/20e5c7742b0a53c16a99d66cd6aec3e03a60a038 and used inside `createSession` function. Now, this part of code is no more present (https://github.com/jenkinsci/jira-plugin/blob/b061f3fee1e3f99aa02b04b842aa98eadc4fd328/src/main/java/hudson/plugins/jira/JiraSite.java#L536-L564), it has been removed here https://github.com/jenkinsci/jira-plugin/commit/85c6b2aa8d8bdd34973d5f46b74b1c2eb3f131a4.

EliaBracciSumo: What  do you think if I move `useHTTPAuth` and `useBearerAuth` in a list with default value to HTTP? Remove this variable means touch too much code IMO. Maybe it should be done in another PR

EliaBracciSumo:
Screenshot 2023-03-30 at 12 26 15
Screenshot 2023-03-30 at 12 26 29
Screenshot 2023-03-30 at 12 26 48

src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
EliaBracciSumo: In the screenshots you can see an example. However, this implementation need further changes since the variable of the select dropdown list should be saved in the `JiraSite` builder and current implementation will change. IDK if the best solution, for now, is to just remove `useHTTPAuth` from `config.jelly` and then open a new PR by removing all code related to `useHTTPAuth` variable.
Screenshot 2023-03-30 at 15 38 01
Screenshot 2023-03-30 at 15 37 58

EliaBracciSumo: I hid temporarily `useHTTPAuth` checkbox. Then, if needed, I can open a new PR and cleanup the code for `useHTTPAuth` variable 👍

EliaBracciSumo:
@rantoniuk I think that the PR can be merged 👍 I'm confident with the changes and I tested it. Just a note, from the previous screen I changed `useBearerAuth` checkbox title from `Use Bearer authentication` to `Use Bearer authentication instead of Basic authentication`

EliaBracciSumo:
Hi @rantoniuk 😄 Any update on that side? Am I missing something?

rantoniuk:
@EliaBracciSumo sorry for the delay. I'd like to test your changes against our Jira Cloud instance if all looks good. Could you briefly explain how would you configure the plugin after your changes?

I took a quick try to test it already by:
- spinning up a clean docker-based LTS instance and installing the plugin [from incrementals](https://repo.jenkins-ci.org/incrementals/org/jenkins-ci/plugins/jira/3.10-rc1460.e4628e65a_a_80/)
- then configured a System credential with a [freshly generated PAT token](https://support.atlassian.com/atlassian-account/docs/manage-api-tokens-for-your-atlassian-account/) 
- then I validated it works by hitting the _Validate Settings_ button with the `Use Bearer authentication instead of Basic authentication` **unchecked** -> the result was success
- then I checked the Bearer checkbox and I tried to validate again - the validation failed

Am I missing something? If you wish to try it out against the Jira Cloud instance, let me know and I will create an account for you for tests.

EliaBracciSumo:
> @EliaBracciSumo sorry for the delay. I'd like to test your changes against our Jira Cloud instance if all looks good. Could you briefly explain how would you configure the plugin after your changes?
> 
> I took a quick try to test it already by:
> 
> * spinning up a clean docker-based LTS instance and installing the plugin [from incrementals](https://repo.jenkins-ci.org/incrementals/org/jenkins-ci/plugins/jira/3.10-rc1460.e4628e65a_a_80/)
> * then configured a System credential with a [freshly generated PAT token](https://support.atlassian.com/atlassian-account/docs/manage-api-tokens-for-your-atlassian-account/)
> * then I validated it works by hitting the _Validate Settings_ button with the `Use Bearer authentication instead of Basic authentication` **unchecked** -> the result was success
> * then I checked the Bearer checkbox and I tried to validate again - the validation failed
> 
> Am I missing something? If you wish to try it out against the Jira Cloud instance, let me know and I will create an account for you for tests.
Bearer credentials with `Use Bearer authentication instead of Basic authentication` **checked**:
Screenshot 2023-05-09 at 15 01 09 
Bearer credentials with `Use Bearer authentication instead of Basic authentication` **unchecked**:
Screenshot 2023-05-09 at 15 02 16
🤔 
I think that create an account for my tests could be a nice idea 👍

EliaBracciSumo:
@rantoniuk have you tried to use PAT with `curl`?

rantoniuk:
> @rantoniuk have you tried to use PAT with curl?

I didn't because I know PAT method worked before with the Basic auth.

> 🤔 I think that create an account for my tests could be a nice idea 👍

Please try to access https://jenkins-jira-plugin.atlassian.net/ with your Atlassian account, you should be able to request access.

> Which versions of the plugin did you install?

See my previous comment, there is a link to incrementals build (so the version built from this PR).

EliaBracciSumo:
@rantoniuk you are right. https://community.atlassian.com/t5/Jira-questions/Jira-on-prem-cloud-API-Personal-Access-Token-PAT/qaq-p/2030384 Jira cloud doesn't allow to create PAT but just API token and this one can be used only with `Basic Auth`. So to recap:
- on-premise Jira: PAT + API token (both Basic and Bearer Auth)
- cloud Jira: API  token (only Basic Auth)

Since my update works fine with on-premise Jira, maybe I can put a not on the checkbox I added saying **Note: Bearer authentication doesn't work with Jira Cloud, use Basic instead.** WDYT?

src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
```diff
@@ -10,7 +10,7 @@
     <f:checkbox title="${%Use HTTP authentication instead of normal login}" field="useHTTPAuth" />
   </f:invisibleEntry>
   <f:entry>
-     <f:checkbox title="${%Use Bearer authentication instead of Basic authentication}" field="useBearerAuth" />
+     <f:checkbox title="${%Use Bearer authentication instead of Basic authentication (Note: Bearer authentication is only supported in Jira Server, for Jira Cloud leave this unchecked)}" field="useBearerAuth" />
```
rantoniuk, line 13: You can use `description=""` and add relevant translation description into [config.properties](https://github.com/jenkinsci/jira-plugin/blob/master/src/main/resources/hudson/plugins/jira/JiraSite/config.properties), see example in line 6.

src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
```diff
@@ -6,8 +6,11 @@
   <f:entry title="Link URL" field="alternativeUrl" description="${%site.alternativeUrl}">
     <f:textbox />
   </f:entry>
-  <f:entry>
+  <f:invisibleEntry>
     <f:checkbox title="${%Use HTTP authentication instead of normal login}" field="useHTTPAuth" />
+  </f:invisibleEntry>
+  <f:entry>
+    <f:checkbox title="${%Use Bearer authentication instead of Basic authentication}" field="useBearerAuth" description="${%Note: Bearer authentication is only supported in Jira Server, for Jira Cloud leave this unchecked}" />
```
rantoniuk, line 13: Almost :-) This should be: `${%site.useBearerAuth}` and then [config.properties](https://github.com/jenkinsci/jira-plugin/blob/master/src/main/resources/hudson/plugins/jira/JiraSite/config.properties):
```
site.useBearerAuth=....
```

EliaBracciSumo: 🤦 Sorry

src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
```diff
@@ -10,7 +10,7 @@
     <f:checkbox title="${%Use HTTP authentication instead of normal login}" field="useHTTPAuth" />
   </f:invisibleEntry>
   <f:entry>
-    <f:checkbox title="${%Use Bearer authentication instead of Basic authentication}" field="useBearerAuth" description="${%Note: Bearer authentication is only supported in Jira Server, for Jira Cloud leave this unchecked}" />
+    <f:checkbox title="${%Use Bearer authentication instead of Basic authentication}" field="useBearerAuth" description="${%site.useBearerAuth}" />
```
rantoniuk, line 13: I just tested the last build but the description text did not show up. So it seems the `description` needs to be on the [<f:entry>](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/lib/form/entry.jelly) tag instead of [<f:checkbox>](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/lib/form/checkbox.jelly). After that change the description showed up for me fine.

src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
rantoniuk: Thanks a lot for this contribution, you did a great job! Fingers crossed 🤞🏻 that you find some time to help with the clean up in the future!

src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
EliaBracciSumo: Should I move just description on entry level or do you think that it's better to move everything (e.g. title and description) on entry level?

src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
EliaBracciSumo: Yeah that could be done for sure 👍 In the meantime, I've another question posted [here](https://github.com/jenkinsci/jira-plugin/pull/145#issuecomment-1500275153)

rantoniuk: Oh, I missed this! Is this blocking this PR from being merged? I don't see `addAttachment` anywhere in this PR, but maybe I missed it somewhere. If that's the case, is there a test that can be added to validate this?

src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
EliaBracciSumo: For some reason my comment hasn't been published.. Should I move just description on entry tag or also field and title attributes?

src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
EliaBracciSumo: No, it's not blocking this PR to be merged 👍 I'll provide you some error logs asap

src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
rantoniuk: Just the description, since [field](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/lib/form/checkbox.jelly#L62) and [title](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/lib/form/checkbox.jelly#L65) are valid for checkbox, only `description` does not exist in that jelly (for unknown reason for me..)

src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
rantoniuk: So if I understand correctly this is an unrelated topic, please open a separate issue for this one.

src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
EliaBracciSumo: Done 👍

src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
EliaBracciSumo: Yeah correct, I just asked on this old PR since I thought that someone already know the response to my question 🤣 I'll open a new issue

src/test/java/hudson/plugins/jira/auth/JiraRestServiceBearerAuthTest.java
```diff
@@ -0,0 +1,63 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.any;
+import static org.mockito.Mockito.anyInt;
+import static org.mockito.Mockito.anyLong;
+import static org.mockito.Mockito.doReturn;
+
+public class JiraRestServiceBearerAuthTest {
```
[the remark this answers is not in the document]
EliaBracciSumo: We discussed [here](https://github.com/jenkinsci/jira-plugin/pull/521#discussion_r1149793625) about `JiraTesterBearerAuth` class: this class is used for manual testing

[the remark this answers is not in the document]
rantoniuk: Right, that was my missing link hidden [in the constructor](https://github.com/jenkinsci/jira-plugin/pull/521/files/03b244067c7da34f12359e3bd9f1366624355e67#diff-547ba84057cbb60c1e431bf068281e854e1f720c952c3337c89647967ff337ffR49), it's not obvious at first glance, thanks!

EliaBracciSumo:
@rantoniuk thank you for all the support!! In the next days I will proceed (ASAP) with the two pending tasks (repo cleaning and missing apache library)

PR #555, opened 2023-06-23
Format code with Spotless

### LINKED ISSUE #553: Enable automated code formatting
Ref: https://github.com/jenkinsci/plugin-pom/pull/733
Depends on bumping plugin-pom to 4.66 via #549

basil:
Fixes #553. Turns on the project-wide Spotless formatting settings and runs `mvn spotless:apply` against this code base. Tested with `mvn clean verify -DskipTests`. If accepted, I will follow up this PR with another PR to add the final commit to `.git-blame-ignore-revs`.


pom.xml
```diff
@@ -58,69 +34,57 @@
...
+    <revision>3.11</revision>
+    <changelist>-SNAPSHOT</changelist>
+    <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
+    <jira-rest-client.version>5.2.5</jira-rest-client.version>
+    <fugue.version>4.7.2</fugue.version>
+    <!-- jenkins -->
+    <jenkins.version>2.375.2</jenkins.version>
+    <!-- security spotbugs -->
+    <spotbugs.failOnError>false</spotbugs.failOnError>
+    <spotless.check.skip>false</spotless.check.skip>
```
rantoniuk, line 59: - should that be changed to true?
- should I also see `.mvn_exec_spotless` in this PR?

rantoniuk:
@basil thanks a lot for this! One question that I don't see an answer for, is how to configure IDE for this? E.g. I'm using VSCode and Prettier, what configuration do I need to do to pick this up automatically so that the Prettier formatting doesn't do something different than the spotless expects?

Can you update [CONTRIBUTING.md](https://github.com/jenkinsci/jira-plugin/blob/master/CONTRIBUTING.md#code-style) accordingly?

pom.xml
basil: > should that be changed to true?

It should not.

> should I also see `.mvn_exec_spotless` in this PR?

You should not.

basil:
> One question that I don't see an answer for, is how to configure IDE for this? E.g. I'm using VSCode and Prettier, what configuration do I need to do to pick this up automatically so that the Prettier formatting doesn't do something different than the spotless expects?

I do not know.

rantoniuk:
That's interesting, I was expecting some more elaborate answer TBH ;-) So, you don't know because you do not use any IDE auto-formatting on save?

pom.xml
rantoniuk: And this is because? Is your "How do I adopt this?" section outdated from https://github.com/jenkinsci/plugin-pom/pull/733?

How about referencing this somewhere on https://www.jenkins.io/doc/developer/development-environment/ide-configuration/?

basil:
I do not use any IDE auto-formatting.

pom.xml
basil: https://github.com/jenkinsci/plugin-pom/pull/733 is outdated. For current documentation, see:

https://github.com/jenkinsci/plugin-pom#formatting

rantoniuk:
I'll go ahead and merge and update the README file accordingly.

PR #557, opened 2023-06-25
Slim down HPI

basil:
Removes these JARs from the HPI:

- `WEB-INF/lib/commons-codec-1.15.jar`
- `WEB-INF/lib/commons-io-2.11.0.jar`
- `WEB-INF/lib/jettison-1.5.4.jar`
- `WEB-INF/lib/jsr305-3.0.1.jar`
- `WEB-INF/lib/slf4j-api-2.0.3.jar`

SLF4J, Commons Codec, and Commons IO are already provided by Jenkins core. Due to how [plugin class loading works in Jenkins](https://www.jenkins.io/doc/developer/plugin-development/dependencies-and-class-loading/), core's copy is always first and the plugin copy is never visible. So it is just a waste of space to bundle it in the HPI, since it is never used.

Jettison is similarly already bundled in the Jersey 2 API plugin, which is on the classpath via a plugin-to-plugin dependency.

JSR 305 is not needed at runtime and is deprecated. Throughout the whole Jenkins project, we have been getting rid of JSR 305 annotations in favor of SpotBugs annotations.

Tested with `mvn clean verify` locally.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -78,11 +78,6 @@
         <artifactId>json</artifactId>
         <version>20230227</version>
       </dependency>
-      <dependency>
-        <groupId>org.springframework</groupId>
-        <artifactId>spring-jcl</artifactId>
-        <version>6.0.9</version>
-      </dependency>
     </dependencies>
   </dependencyManagement>
 
@@ -92,14 +87,15 @@
       <artifactId>jira-rest-java-client-api</artifactId>
       <version>${jira-rest-client.version}</version>
       <exclusions>
+        <!-- Deprecated and not needed at runtime -->
+        <exclusion>
+          <groupId>com.google.code.findbugs</groupId>
+          <artifactId>jsr305</artifactId>
+        </exclusion>
+        <!-- Provided by Jenkins core -->
         <exclusion>
           <groupId>com.google.guava</groupId>
           <artifactId>guava</artifactId>
-        </exclusion>
-        <!-- provided by commons-lang3-api -->
-        <exclusion>
-          <groupId>org.apache.commons</groupId>
-          <artifactId>commons-lang3</artifactId>
         </exclusion>
       </exclusions>
     </dependency>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -78,11 +78,6 @@
         <artifactId>json</artifactId>
         <version>20230227</version>
       </dependency>
-      <dependency>
-        <groupId>org.springframework</groupId>
-        <artifactId>spring-jcl</artifactId>
-        <version>6.0.9</version>
-      </dependency>
```
basil, line 85: Recent versions of the Jenkins core BOM (not to be confused with the Jenkins plugin BOM), which is imported in the Jenkins plugin parent POM, import the Spring BOM. So there is no need to manage Spring versions anywhere except in Jenkins core.

pom.xml
```diff
@@ -92,14 +87,15 @@
...
-          <groupId>com.google.guava</groupId>
-          <artifactId>guava</artifactId>
+          <groupId>com.google.code.findbugs</groupId>
+          <artifactId>jsr305</artifactId>
         </exclusion>
-        <!-- provided by commons-lang3-api -->
+        <!-- Provided by Jenkins core -->
         <exclusion>
-          <groupId>org.apache.commons</groupId>
-          <artifactId>commons-lang3</artifactId>
```
basil, line 102: `commons-lang3` is not a dependency of this artifact, so there is no need to exclude it.

pom.xml
```diff
@@ -116,28 +112,59 @@
...
+          <artifactId>jsr305</artifactId>
+        </exclusion>
+        <!-- Provided by Jenkins core -->
         <exclusion>
           <groupId>com.google.guava</groupId>
           <artifactId>guava</artifactId>
         </exclusion>
         <exclusion>
-          <groupId>javax.activation</groupId>
-          <artifactId>activation</artifactId>
```
basil, line 125: `activation` is not a dependency of this artifact, so there is no need to exclude it.

pom.xml
```diff
@@ -154,17 +181,25 @@
...
+          <groupId>org.springframework</groupId>
+          <artifactId>spring-jcl</artifactId>
+        </exclusion>
       </exclusions>
     </dependency>
-    <dependency>
-      <groupId>commons-io</groupId>
-      <artifactId>commons-io</artifactId>
-      <version>2.12.0</version>
-    </dependency>
```
basil, line 167: `commons-io` is already on the classpath by virtue of the dependency of Jenkins core. It makes no sense to manage our own version here or to bundle Commons IO in the HPI, because core's copy will always be loaded by the Jenkins plugin class loading system.

pom.xml
```diff
@@ -182,7 +217,6 @@
     <dependency>
       <groupId>io.jenkins.plugins</groupId>
       <artifactId>commons-lang3-api</artifactId>
-      <version>3.12.0-36.vd97de6465d5b_</version>
```
basil, line 185: Using the version managed in the Jenkins plugin BOM.

pom.xml
```diff
@@ -248,30 +283,17 @@
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <artifactId>workflow-step-api</artifactId>
       <optional>true</optional>
-      <exclusions>
-        <exclusion>
-          <groupId>org.jenkins-ci.plugins</groupId>
-          <artifactId>*</artifactId>
-        </exclusion>
-      </exclusions>
```
basil, line 256: Unnecessary as of the latest plugin BOM.

pom.xml
```diff
@@ -248,30 +283,17 @@
...
     <dependency>
       <groupId>io.jenkins.configuration-as-code</groupId>
       <artifactId>test-harness</artifactId>
       <scope>test</scope>
-      <exclusions>
-        <exclusion>
-          <groupId>joda-time</groupId>
-          <artifactId>joda-time</artifactId>
-        </exclusion>
-      </exclusions>
```
basil, line 267: Unnecessary as of the latest plugin BOM.

pom.xml
```diff
@@ -248,30 +283,17 @@
...
     </dependency>
     <!-- test dependencies -->
     <!-- required to start p4 tests -->
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>command-launcher</artifactId>
-      <version>100.v2f6722292ee8</version>
```
basil, line 274: Using the version managed in the Jenkins plugin BOM.

rantoniuk:
That's an interesting PR. I have tested this and all indeed looks correct. My question is, how can a plugin developer detect (preferably automatically) a dependency that is managed via jenkins-bom or plugin-pom?
I'm guessing you're just doing some kind of analysis of `dependency:tree`, just wondering if that can be somehow plugged into `verify` phase.

A good example is `joda-time`, obviously noone has time to follow all new releases there, so I'm curious if we can detect duplicates easily when updating to the newer bom version via dependabot?

rantoniuk (approved):

pom.xml
rantoniuk: Are all instances of this unnecessary, i.e. [this one as well](https://github.com/jenkinsci/jira-plugin/blob/master/pom.xml#L274)?

basil:
Some people have talked about using OpenRewrite to automatically replace dependencies on libraries with dependencies on library wrapper Jenkins plugins to implement dynamic linking. Joda Time, like other common libraries, ought to be wrapped in a library wrapper Jenkins plugin rather than embedded into the HPI of every consumer in the Jenkins plugin ecosystem, but it appears that nobody has had the time or energy to do that.

pom.xml
basil: Not sure offhand.

TobiX:
> Joda Time, like other common libraries, ought to be wrapped in a library wrapper Jenkins plugin rather than embedded into the HPI of every consumer in the Jenkins plugin ecosystem, but it appears that nobody has had the time or energy to do that.

I put that energy into removing Joda from my own plugins, which probably all-in-all will take more time since migrating from Joda to JSR-310 is not always straightforward unfortunately...

PR #570, opened 2023-09-08
[JENKINS-71957] Fix support for Java 21

alecharp:
# Description
See [JENKINS-71957](https://issues.jenkins.io/browse/JENKINS-71957).

When building the plugin with Java 17 and less, everything is fine but once building with Java 20+, there is a NPE occurring in the test.

This is because when testing with invalid setup, the exception is supposed to be printed out. However, the way the exceptions are printed changed. More specifically, the way a lock is acquired in order to print the exceptions. This should have no consequences on any code, as it is internal change of the JVM, but in the current plugin it creates a NPE are the object used to get the lock is a `mock` which is not properly initialized in the test class.

# Testing

I ran `mvn clean verify` with Java 17 an no issue happened.
I ran `mvn clean verify` with both Java 20 and 21 and then both `JiraCreateReleaseNotesTest.failBuildOnErrorEmptyProjectKey` and `JiraCreateReleaseNotesTest.failBuildOnErrorEmptyRelease` failed with 

```
NullPointer Cannot enter synchronized block because "lock" is null
```

With this changest, this is not occuring anymore.

To run this and validate the changes, the new `Jenkinsfile` needs to be used, so a maintainer or an admin must re-open the pull request.

### DIFF
--- a/src/test/java/hudson/plugins/jira/JiraCreateReleaseNotesTest.java
+++ b/src/test/java/hudson/plugins/jira/JiraCreateReleaseNotesTest.java
@@ -21,6 +21,7 @@ import hudson.model.Item;
 import hudson.model.Result;
 import hudson.tasks.BuildWrapper;
 import java.io.IOException;
+import java.io.OutputStream;
 import java.io.PrintWriter;
 import java.util.Arrays;
 import java.util.HashMap;
@@ -62,8 +63,7 @@ public class JiraCreateReleaseNotesTest {
     @Mock
     JiraSite site;
 
-    @Mock
-    private PrintWriter printWriter;
+    private final PrintWriter printWriter = new PrintWriter(OutputStream.nullOutputStream());
 
     @Mock
     JiraSession session;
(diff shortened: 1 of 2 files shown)


src/test/java/hudson/plugins/jira/JiraCreateReleaseNotesTest.java
```diff
@@ -62,8 +63,7 @@ public class JiraCreateReleaseNotesTest {
     @Mock
     JiraSite site;
 
-    @Mock
-    private PrintWriter printWriter;
+    private final PrintWriter printWriter = new PrintWriter(OutputStream.nullOutputStream());
```
basil, line 66: [nit] The only real reason to make this a field in the first place was to use the `@Mock` annotation on it, so now that `@Mock` is being deleted this could be inlined and would be arguably more readable that way. But I do not feel strongly either way.

basil (approved):

alecharp:
Thanks @rantoniuk. Do you think you could release the plugin? Thanks

PR #578, opened 2023-12-04
Explicitly use Atlassian repository

basil:
See https://github.com/jenkins-infra/helpdesk/issues/3842. This plugin currently depends on JCenter for Atlassian libraries, but JFrog has asked us to stop caching JCenter dependencies. Based on [this page](https://developer.atlassian.com/server/framework/atlassian-sdk/working-with-maven/), the dependencies we need are officially published at https://packages.atlassian.com/maven-public/ so fetch them from there directly. We choose `maven-public` rather than `maven-external` because Atlassian may one day go through a similar bandwidth reduction exercise as we are currently going through right now, and in such an event they would want us to rely on public repositories for third-party libraries, so proactively anticipate such an event by using the repository that contains just Atlassian libraries.

To test this, I verified that I could run `mvn clean verify -DskipTests` in a clean Docker container (without `~/.m2`) and without the JCenter cache:

```xml
  <repositories>
    <repository>
      <releases>
        <enabled>true</enabled>
      </releases>
      <snapshots>
        <enabled>false</enabled>
      </snapshots>
      <id>repo.jenkins-ci.org</id>
      <url>https://repo.jenkins-ci.org/releases/</url>
    </repository>
    <repository>
      <releases>
        <enabled>true</enabled>
      </releases>
      <snapshots>
        <enabled>false</enabled>
      </snapshots>
      <id>orphans</id>
      <url>https://repo.jenkins-ci.org/artifactory/orphans/</url>
    </repository>
    <repository>
      <releases>
        <enabled>true</enabled>
      </releases>
      <snapshots>
        <enabled>false</enabled>
      </snapshots>
      <id>atlassian-maven-public</id>
      <url>https://packages.atlassian.com/maven-public/</url>
    </repository>
  </repositories>
```

Note that the `orphans` repository is needed for `jbcrypt` as explained in https://github.com/jenkins-infra/helpdesk/issues/3842.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -330,6 +330,16 @@
       <id>repo.jenkins-ci.org</id>
       <url>https://repo.jenkins-ci.org/public/</url>
     </repository>
+    <repository>
+      <releases>
+        <enabled>true</enabled>
+      </releases>
+      <snapshots>
+        <enabled>false</enabled>
+      </snapshots>
+      <id>atlassian-maven-public</id>
+      <url>https://packages.atlassian.com/maven-public/</url>
+    </repository>
   </repositories>
 
   <pluginRepositories>


MarkEWaite (approved):
Thanks!

basil (commented):
In draft pending a decision as to whether we want to create a remote repository for https://packages.atlassian.com/maven-public/ and include it in the https://repo.jenkins-ci.org/public/ virtual repository as described in https://github.com/jenkins-infra/helpdesk/issues/3842#issuecomment-1839659393.

rantoniuk:
I remembered this ringed a bell, [we've been there before](https://github.com/jenkinsci/jira-plugin/commit/2a7d5f12692c36b2bdc50f01960a460e313faddf).

MarkEWaite:
I think we should add the Atlassian repository to the pom.xml of this plugin rather than mirror the contents of the Atlassian repository in the Jenkins repository because we're trying to reduce the bandwidth used by the Jenkins repository.  Adding another mirror seems like it will increase the bandwidth use rather than reduce it.

rantoniuk:
Well, https://github.com/jenkinsci/jira-plugin/pull/471

PR #579, opened 2023-12-11
Implement dynamic linking of Joda Time

basil:
Stop bundling an unnecessary copy of Joda Time; instead, implement dynamic linking by depending on @jonesbusy's new [Joda Time](https://plugins.jenkins.io/joda-time-api/) library plugin. This, in turn, results in updating the baseline (to use the newest BOM release where the Joda Time library plugin is present), but this bump is in accordance with [the official recommendations](https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline/).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -53,7 +53,7 @@
     <jira-rest-client.version>5.2.7</jira-rest-client.version>
     <fugue.version>4.7.2</fugue.version>
     <!-- jenkins -->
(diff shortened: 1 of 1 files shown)


basil:
Thanks! Looking forward to the release of this PR so that we can complete the migration to the Joda Time library plugin.

PR #598, opened 2024-05-08
Replace JSR-305 annotations with spotbugs annotations

MarkEWaite:
## Replace JSR-305 annotations with spotbugs annotations

Annotations for Nonnull, CheckForNull, and several others were proposed for Java as part of dormant Java specification request JSR-305. The proposal never became a part of standard Java.

Jenkins plugins should switch from using JSR-305 annotations to use Spotbugs annotations that provide the same semantics.

The [mailing list discussion](https://groups.google.com/g/jenkinsci-dev/c/uE1wwtVi1W0/m/gLxdEJmlBQAJ) from James Nord describes the affected annotations and why they should be replaced with annotations that are actively maintained.

The ["Improve a plugin" tutorial](https://www.jenkins.io/doc/developer/tutorial-improve/replace-jsr-305-annotations/) provides instructions to perform this change.

An [OpenRewrite recipe](https://docs.openrewrite.org/recipes/jenkins/javaxannotationstospotbugs) is also available and is even better than the tutorial.

Confirmed that automated tests pass on Linux with Java 21.

### DIFF
--- a/src/main/java/hudson/plugins/jira/JiraSite.java
+++ b/src/main/java/hudson/plugins/jira/JiraSite.java
@@ -73,7 +73,6 @@ import java.util.concurrent.locks.ReentrantLock;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import java.util.regex.Pattern;
-import javax.annotation.Nonnull;
 import javax.annotation.PreDestroy;
 import javax.servlet.ServletException;
 import jenkins.model.Jenkins;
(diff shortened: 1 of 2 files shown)


rantoniuk:
Thanks! I'm curious, can we exclude jsr305 completely and/or add a spotbugs check to cause an error if someone tries to reintroduce non-compliant annotations again?

MarkEWaite:
> Thanks! I'm curious, can we exclude jsr305 completely (probably in bom/plugin-pom) and/or add a spotbugs check to cause an error if someone tries to reintroduce non-compliant annotations again?

That's a very good idea.  As  more and more plugins move away from JSR-305 annotations, we should be able to eventually exclude them in the plugin parent pom and the Jenkins core parent pom.  I submitted pull requests for about 20 of the most popular 250 plugins but did not yet search more widely than that.

MarkEWaite:
> How do you want to proceed? Do you want to add that exclusion here or should we merge this now and you can add an exclusion or SpotBugs check in the next version of parent pom?

I'd prefer that this be merged as it is, without the exclusion.  I think that the risk of insertion of a new JSR-305 annotation into this repository is very low and removing the exclusion in the future will be a TODO for this plugin without much benefit to the plugin maintainers.

Including the exclusion in the plugin parent pom and the Jenkins core parent pom will need more discussion and will probably need more validation that the consumers are generally ready before we add the exclusion.

PR #599, opened 2024-05-08
Fix/suppress spotbugs warnings

MarkEWaite:
## Fix / suppress spotbugs warnings

Fixes warnings with code changes in:

- 010be1f0b9e71e76d34648226387263b74dd1b7b **Use %n instead of \n in formatting strings**
- c26ca8f585e275414b72cee32376455dc0374ca4 **Conditional already checked not null**
- 71ce36f78f6e09eb38051e1fe53edd104e5fd0da **Remove redundant null check in Updater**

Switches to require spotbugs to remain warning free.

- 61b0bc8083a3296acc676df8c2748554a91d01a3 **Fail if new spotbugs errors are introduced**

If it is preferred to only suppress the warnings that have been triaged, I'm happy to do that instead.  I think that it is helpful to have the safety of not introducing new spotbugs warnings by failing the build when a new spotbugs warning is introduced.

### DIFF
--- a/src/spotbugs/excludesFilter.xml
+++ b/src/spotbugs/excludesFilter.xml
@@ -1,0 +1,65 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<FindBugsFilter>
+  <!--
+    Exclusions in this section have been triaged and determined to be
+    false positives.
+  -->
+  <Match>
+    <!-- Message string names are wrong style for Java but that's how Jenkins does it -->
+    <Bug pattern="NM_METHOD_NAMING_CONVENTION" />
+    <Class name="hudson.plugins.jira.Messages" />
+    <Or>
+      <Method name="ErrorCommentingIssues" />
+      <Method name="FailedToConnect" />
+      <Method name="FailedToUpdateIssue" />
+      <Method name="FailedToUpdateIssueWithCarryOver" />
+      <Method name="NoJenkinsUrl" />
+      <Method name="NoJiraSite" />
+      <Method name="NoRemoteAccess" />
+      <Method name="UpdatingIssue" />
+    </Or>
+  </Match>
+  <Match>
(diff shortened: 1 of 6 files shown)


src/spotbugs/excludesFilter.xml
```diff
@@ -0,0 +1,65 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<FindBugsFilter>
+  <!--
+    Exclusions in this section have been triaged and determined to be
+    false positives.
+  -->
+  <Match>
```
rantoniuk, line 7: Can an line `@SuppressFBWarnings` on the method or class be used instead directly in the source file rather than having a separate XML file? I prefer to keep all things in one place.

rantoniuk (approved):

src/spotbugs/excludesFilter.xml
MarkEWaite: Yes, individual suppressions can be used, though that will scatter many more suppressions in the source files.  In particular, the `NM_METHOD_NAMING_CONVENTION` suppression will need to find the right location to insert that suppression for variable names without suppressing all checks for inconsistently named variables.

I've had a mixed experience with inline suppression of spotbugs warnings.  When I want to provide a detailed reason for a method or field level suppression, inline suppressions have worked reasonably well.  When I wanted a more general suppression, the XML configuration file seemed to work better as a central location to configure suppressions.  Both techniques work and each has their own strengths and weaknesses.

I'm not sure that I'll have time to move from centrally defined suppressions to suppressions inside the individual files for this repository.  This pull request was a side effect of #598 and already took more time than I had planned.  I'm not sure that I have additional time to spend on this side effect.  If you'd like to make that change, you're welcome to do it.  You're welcome to use any or all of this pull request as a basis for the change.  Thanks for maintaining the Jira plugin!

PR #613, opened 2024-08-20
[JENKINS-73137] Adapt Jira plugin for jetty 12 EE8

BorisYaoA:
see https://issues.jenkins.io/browse/JENKINS-73137

The goal of this PR is to adapt Jira-plugin for jetty 12 (EE8).
The most of the work is about adapting `ProxyTestHandler` and `TestHandler` classes to use Handler.Abstract from jetty12

- [X] Make sure you are requesting to **pull a topic/feature/bugfix branch** (right side) and not your master branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did
- [X] Link to relevant issues in GitHub 
- [X] Link to relevant pull requests, esp. upstream and downstream changes
- [X] Ensure you have provided tests - that demonstrates feature works or fixes the issue


pom.xml
```diff
@@ -52,17 +52,22 @@
...
     <spotless.check.skip>false</spotless.check.skip>
   </properties>
 
   <dependencyManagement>
     <dependencies>
       <dependency>
         <groupId>io.jenkins.tools.bom</groupId>
-        <artifactId>bom-2.440.x</artifactId>
-        <version>3234.v5ca_5154341ef</version>
+        <artifactId>bom-${jenkins.baseline}.x</artifactId>
```
alecharp: why no just keeping `bom-2.462.x` here? Why adding a property on the pom?

basil (approved):
I have replayed the CI build with the `Jenkinsfile` changes here: https://ci.jenkins.io/job/Plugins/job/jira-plugin/job/PR-613/10/

For the benefit of the plugin maintainers, I wanted to explain why this PR raises the baseline to a weekly release version, which is atypical as described in https://www.jenkins.io/doc/developer/plugin-development/choosing-jenkins-baseline and elsewhere. We generally try to preserve compatibility as much as possible, but in this case the incompatible change was made upstream in Jetty. We generally try to test with the same version of Jetty as is used in production. As of recent weeklies, that is Jetty 12, but Jetty 12 requires compilation changes to tests.

While this change could be postponed in order to keep the baseline low, it needs to be made eventually. It might be simplest to get it out of the way now to prevent bitrot and improve development efficiency. It does not have to be released right away. It could be released in a few months, once Jetty 12 is in LTS.

src/test/java/com/atlassian/httpclient/apache/httpcomponents/ApacheAsyncHttpClientTest.java
```diff
@@ -164,55 +163,50 @@ public class ProxyTestHandler extends AbstractHandler {
         final String realm = "test_realm";
 
         @Override
-        public void handle(
-                String target,
-                org.eclipse.jetty.server.Request jettyRequest,
-                HttpServletRequest request,
-                HttpServletResponse response)
-                throws IOException, ServletException {
+        public boolean handle(
```
olamy, line 166: changes looks good. 
But I wonder if it would be better to simply use Servlet API filter to have some implementation of this not depending on Jetty core API (which may change soon again for 12.1.x)

Jenkinsfile
```diff
@@ -1,7 +1,6 @@
 // Builds a module using https://github.com/jenkins-infra/pipeline-library
 buildPlugin(useContainerAgent: true, configurations: [
-    [platform: 'linux', jdk: 11],
```
rantoniuk, line 3: We're not dropping JDK11 until there is an LTS release that is no longer supporting Java11.

pom.xml
```diff
@@ -52,17 +52,20 @@
     <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
     <jira-rest-client.version>5.2.7</jira-rest-client.version>
     <fugue.version>4.7.2</fugue.version>
-    <!-- jenkins -->
-    <jenkins.version>2.440.3</jenkins.version>
+    <!-- TODO: Remove when 2.472 is available in LTS -->
+    <jenkins.version>2.472</jenkins.version>
```
rantoniuk, line 56: We are supporting only LTS line, so let's make this a draft until there is an LTS released >=2.472
https://www.jenkins.io/changelog-stable/

pom.xml
```diff
@@ -52,17 +52,20 @@
...
+    <!-- TODO JENKINS-73339 until in parent POM -->
+    <jenkins-test-harness.version>2254.vcff7a_d4969e5</jenkins-test-harness.version>
     <spotless.check.skip>false</spotless.check.skip>
+    <maven.compiler.release>17</maven.compiler.release>
```
rantoniuk, line 60: This is/should be coming from parent-pom, we don't want an override here

rantoniuk (changes requested):

basil:
Hi @rantoniuk, that would imply that this PR would need to sit around for a few more months before it could be merged, by which point the staffing for the current initiative would have ended, and the author of the PR may not necessarily be available to rebase the PR, deal with merge conflicts, address review feedback, etc. So I think there are two options:

- Accept this PR now in its current form, but delay a release of the Jira plugin until the October LTS release of Jenkins is available. This takes advantage of current staffing and the current state of this PR (passing CI build, no conflicts, etc)
- Block this PR and accept that since there may not be staffing to complete it later you will need to complete the work later yourself

pom.xml
basil: @rantoniuk There is not yet a release of plugin parent POM that requires Java 17, but there will be one around the October timeframe, at which point this temporary code can be deleted. We aren't releasing a plugin parent POM that requires Java 17 right now because most of the ecosystem is still targeting Jenkins core releases that support Java 11.

rantoniuk:
> Accept this PR now in its current form, but delay a release of the Jira plugin until the October LTS release of Jenkins is available. This takes advantage of current staffing and the current state of this PR (passing CI build, no conflicts, etc)

That's not an option, since this would conflict with the agreed LTS-only support policy. If we would merge this PR today, someone would need to do another PR later to fix/revert the non-LTS compatible changes - and that's why I would prefert this stays as a completed draft (i.e. green build) as it will easier to pick it up and update only the properties accordingly when new LTS line is available.

basil:
> That's not an option

Understood, @rantoniuk. Does that imply you are accepting the second option; namely, to complete this work yourself at such a time as it is available in LTS?

pom.xml
rantoniuk: This should be now available in [plugin-pom v.5.4](https://github.com/jenkinsci/plugin-pom/releases) and with [2.479.x](https://github.com/jenkinsci/bom/tree/master/bom-2.479.x) line, right?

pom.xml
rantoniuk: Latest [LTS dropped support for Java11](https://github.com/jenkinsci/bom/tree/master/bom-2.479.x) so after resolving conflicts and updating the jenkins.version line to 2.479.x this should be good to go.

PR #624, opened 2025-01-12
Bump LTS to 2.479.x line

rantoniuk:
JDK 17 and bump LTS to 2.479.x line:
- include https://github.com/jenkinsci/jira-plugin/pull/621 with updates to newest plugin version
- include https://github.com/jenkinsci/jira-plugin/pull/613 with updates for EE9
- include https://github.com/jenkinsci/jira-plugin/pull/625

I cleansed the git history in order not to pollute the master branch (while not using squashing to preserve attribution of the other merged PR).


pom.xml
```diff
@@ -74,7 +74,7 @@
       <dependency>
         <groupId>org.json</groupId>
         <artifactId>json</artifactId>
-        <version>20240303</version>
+        <version>20241224</version>
```
jonesbusy, line 77: Better to use json-api plugin and exclude transitive dependencies that bring org.json

rantoniuk: Good point, maybe that should be added to enforcer rules for all of the existing wrapper plugins?

PR #635, opened 2025-01-22
Migrate from EE 8 to EE 9

basil:
Migrate from deprecated EE 8 APIs to non-deprecated EE 9 equivalents. Tested with `mvn clean verify`.


src/main/java/hudson/plugins/jira/JiraCreateIssueNotifier.java
```diff
@@ -39,7 +39,7 @@
 import org.kohsuke.stapler.AncestorInPath;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.QueryParameter;
-import org.kohsuke.stapler.StaplerRequest;
+import org.kohsuke.stapler.StaplerRequest2;
```
rantoniuk, line 42: I thought the times of having `2` at the end of things is long gone 😆 
How about moving this class to `org.kohsuke.stapler.ee9.StaplerRequest`?

basil: This shipped in Jenkins core months ago and is now part of the public API which cannot be changed.

alecharp:
@rantoniuk Do you think this could be released?
Why? Because, some code was added in PCT to support both EE 8 and EE 9 and that code is not making PCT too complex for new development. Having this release means that the code can be removed. As this plugin is part of the BOM, we cannot remove it until the plugin is released. Thanks.

olamy:
@alecharp, I can look at it as well.

alecharp:
@olamy thank you!

olamy:
@alecharp done

PR #640, opened 2025-03-26
Convert test to `com.sun.net.httpserver`

basil:
When testing this plugin against EE 10, I found the same test-only (not production) issues in `GitHubServerConfigIntegrationTest` as I did when testing this plugin against EE 9. Rather than continue to adapt this test for every EE version, this PR removes any references to Jetty or Jakarta EE from the test, making the test less fragile. In their place, we substitute references to standard Java Platform functionality.

### Testing done

`mvn clean verify` with Java 21


src/test/java/hudson/plugins/jira/JiraSiteSecurity1029Test.java
```diff
@@ -26,11 +30,6 @@
 import jenkins.model.Jenkins;
 import jenkins.security.ApiTokenProperty;
 import net.sf.json.JSONObject;
-import org.eclipse.jetty.ee9.servlet.DefaultServlet;
-import org.eclipse.jetty.ee9.servlet.ServletContextHandler;
-import org.eclipse.jetty.ee9.servlet.ServletHolder;
```
basil, line 31: No more references to a specific EE version, making this test robust against future EE upgrades (including EE 10).

Issue #658, opened 2025-05-22
JIRA Plugin fails in Post-Build-Action "Move issues matching JQL to the specified version"


steingebein:
### Jenkins and plugins versions report

<details>
  <summary>Environment</summary>

  
  ```text
Jenkins: 2.504.1
OS: Linux - 4.18.0-553.51.1.el8_10.x86_64
Jira Plugin Version: 3.15
  ```

</details>

### What Operating System are you using (both controller, and any agents involved in the problem)?

Linux

### Reproduction steps

1.  Create a Jira Release 'APP-NEXT' in project 'APP' and add a fixed issue

2. Create a freestyle Jenkins job.

3. Add Post Build Action "Jira: Create new version"
Parameter 
* Jira Version: APP-1.2.3
* Jira Project Key: APP

4.   Add Post Build Action "Jira: Move issues matching JQL to the specified version"

Parameter
* Target Jira Release: APP-1.2.3
* Add Target Release: false
* Replace Jira Release: APP-NEXT
* Jira Project Key: APP
* JQL Query: project=APP and fixVersion=APP-NEXT

5. Run the job

### Expected Results

the job should create a Jira release 'APP-1.2.3' and move all issues from Jira release 'APP-NEXT' to 'APP-123'. 

### Actual Results

Build Job fails with NullpointerException

[Jira] Creating version APP-1.2.3 in project APP.
FATAL: Unable to release jira version  APP-1.2.3/APP: java.lang.NullPointerException: Cannot invoke "java.lang.Iterable.iterator()" because the return value of "com.atlassian.jira.rest.client.api.domain.Issue.getFixVersions()" is null
java.lang.NullPointerException: Cannot invoke "java.lang.Iterable.iterator()" because the return value of "com.atlassian.jira.rest.client.api.domain.Issue.getFixVersions()" is null
	at PluginClassLoader for jira//hudson.plugins.jira.JiraSession.replaceFixVersion(JiraSession.java:281)
	at PluginClassLoader for jira//hudson.plugins.jira.JiraSite.replaceFixVersion(JiraSite.java:1215)
	at PluginClassLoader for jira//hudson.plugins.jira.JiraIssueMigrator.perform(JiraIssueMigrator.java:121)
	at hudson.tasks.BuildStepMonitor$1.perform(BuildStepMonitor.java:20)
	at hudson.model.AbstractBuild$AbstractBuildExecution.perform(AbstractBuild.java:818)
[...]

PR #661, opened 2025-05-23
Check for null result of getFixVersion

davedevretro:
Issue https://github.com/jenkinsci/jira-plugin/issues/658

Check for null result of getFixVersions as per example on atlassian site:
https://developer.atlassian.com/server/jira/platform/creating-workflow-extensions/#step-2.-edit-the-code

Tested on my local Jenkins instance.

### DIFF
--- a/src/main/java/hudson/plugins/jira/JiraSession.java
+++ b/src/main/java/hudson/plugins/jira/JiraSession.java
@@ -274,16 +274,23 @@ public class JiraSession {
                 LOGGER.fine("Using regular expression: " + regEx);
 
                 Pattern fromVersionPattern = Pattern.compile(regEx);
-                for (Version currentVersion : issue.getFixVersions()) {
-                    Matcher versionToRemove = fromVersionPattern.matcher(currentVersion.getName());
-                    if (!versionToRemove.matches()) {
-                        newVersions.add(currentVersion);
+
+                Iterable<Version> versions = issue.getFixVersions();
+                if(versions != null) {
+                    for (Version currentVersion : issue.getFixVersions()) {
+                        Matcher versionToRemove = fromVersionPattern.matcher(currentVersion.getName());
+                        if (!versionToRemove.matches()) {
+                            newVersions.add(currentVersion);
+                        }
                     }
                 }
             } else {
-                for (Version currentVersion : issue.getFixVersions()) {
-                    if (!currentVersion.getName().equals(fromVersion)) {
-                        newVersions.add(currentVersion);
+                Iterable<Version> versions = issue.getFixVersions();
+                if(versions != null) {
+                    for (Version currentVersion : issue.getFixVersions()) {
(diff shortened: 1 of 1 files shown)


rantoniuk:
Thanks for the contribution! I see you have checked `Ensure you have provided tests` but I don't see any test file modified, did you miss adding it to the PR?

src/main/java/hudson/plugins/jira/JiraSession.java
```diff
@@ -274,16 +274,23 @@ public void replaceFixVersion(String projectKey, String fromVersion, String toVe
...
 
                 Pattern fromVersionPattern = Pattern.compile(regEx);
-                for (Version currentVersion : issue.getFixVersions()) {
-                    Matcher versionToRemove = fromVersionPattern.matcher(currentVersion.getName());
-                    if (!versionToRemove.matches()) {
-                        newVersions.add(currentVersion);
+
+                Iterable<Version> versions = issue.getFixVersions();
+                if(versions != null) {
+                    for (Version currentVersion : versions) {
```
rantoniuk, line 280: How about using [ListUtils.emptyIfNull()](https://commons.apache.org/proper/commons-collections/javadocs/api-4.2/org/apache/commons/collections4/ListUtils.html) instead of adding the `if` clause?

davedevretro:
Sorry I thought this meant that I was confirming it was tested - unticked now.

rantoniuk (changes requested):
Build fails because of format check failure.
You need to run `mvn spotless:apply` after all changes, before committing changes

src/main/java/hudson/plugins/jira/JiraSession.java
rantoniuk: Scratch that, we're stuck on version 3 for now, that doesn't have this.

rantoniuk: Found another way using Java8 Optional:
```java
for (Version currentVersion : Optional.ofNullable(issue.getFixVersions()).orElse(Collections.emptyList())) {
```

davedevretro: Is it starting to break readability for minimal gain at this point?

rantoniuk: I think it looks cleaner than adding another nested `if` statement, especially when there is no else (maybe we should consider adding `LOGGING.fine()`) , but it's fine if you want to leave it as it is.

Would you be able to add a test for this? I think this should be pretty easy by mocking the Null response object.

davedevretro:
updated formatting

src/main/java/hudson/plugins/jira/JiraSession.java
davedevretro: Couldn't work out how to get a unit test to work for this case - sorry.. :/

PR #665, opened 2025-05-29
Add Jira site property to configure max number of returned Jira issues

### LINKED ISSUE #662: Configurability of number of returned issues
### What feature do you want to see added?

Instead of using [a hardcoded value of 100 issues](url) as maxResults default value at getIssuesFromJqlSearch, it'd be great if that value can be configured as part of jira site global configuration.

### Upstream changes

_No response_

Hardikrathod01:
Fixes #662 

Tested on my local Jenkins instance. Have also modified the effected existing test cases.


src/main/resources/hudson/plugins/jira/JiraSite/config.jelly
```diff
@@ -51,6 +51,9 @@
   <f:entry title="${%Jira comments timestamp format}" field="dateTimePattern">
     <f:textbox />
   </f:entry>
+  <f:entry title="${%Max Issues From Jql Search}" field="maxIssuesFromJqlSearch">
+    <f:number default="100" />
```
rantoniuk, line 55: There needs to be a check for maximum allowed value (5000).

rantoniuk (changes requested):

rantoniuk:
> Tested on my local Jenkins instance. Have also modified the effected existing test cases.

With which Jira version and hosting have you tested this (self-hosted/cloud)?

src/main/java/hudson/plugins/jira/JiraSite.java
```diff
@@ -117,6 +117,8 @@ public class JiraSite extends AbstractDescribableImpl<JiraSite> {
 
     public static final int DEFAULT_THREAD_EXECUTOR_NUMBER = 10;
 
+    public static final int DEFAULT_MAX_ISSUES = 100;
```
rantoniuk, line 120: The default comes from Jelly so this shouldn't be needed I guess.

src/main/java/hudson/plugins/jira/JiraSite.java
```diff
@@ -922,16 +949,20 @@ public void destroy() throws Exception {
 
     private static class NoOpEventPublisher implements EventPublisher {
         @Override
-        public void publish(Object o) {}
+        public void publish(Object o) {
```
rantoniuk, line 952: don't reformat unrelated code as this makes the review harder and make sure to run `spotless:apply`

src/main/resources/hudson/plugins/jira/Messages.properties
```diff
@@ -47,3 +47,4 @@ ErrorCommentingIssues=[Jira] Could not comment on some issues: {0}
 JiraSite.threadExecutorMinimunSize = Thread Executor Size must be at least {0} (higher values are recommended)
 JiraSite.timeoutMinimunValue = Connection timeout must be at least {0}
 JiraSite.readTimeoutMinimunValue = Read timeout must be at least {0}
+JiraSite.maxIssuesFromJqlMaximumNumber = Maximum number of allowed issues from jql search are {0}
```
rantoniuk, line 50:
```suggestion
JiraSite.maxIssuesFromJqlMaximumNumber = Maximum number of allowed issues from jql search is {0}
```

src/test/java/JiraTester.java
```diff
@@ -102,8 +102,8 @@ public static void main(String[] args) throws Exception {
 
     private static void callUniq(final JiraRestService restService) throws Exception {
         long start = System.currentTimeMillis();
-        List<Issue> issues =
-                restService.getIssuesFromJqlSearch("key in ('JENKINS-53320','JENKINS-51057')", JiraSession.MAX_ISSUES);
+        List<Issue> issues = restService.getIssuesFromJqlSearch(
+                "key in ('JENKINS-53320','JENKINS-51057')", 100);
```
rantoniuk: this was testing MAX_ISSUES, not the default, so use the `MAX_ALLOWED_ISSUES_FROM_JQL` that you introduced

src/main/java/hudson/plugins/jira/JiraSite.java
```diff
@@ -1350,6 +1374,7 @@ public FormValidation doValidate(
                 @QueryParameter int readTimeout,
                 @QueryParameter int threadExecutorNumber,
                 @QueryParameter boolean useBearerAuth,
+                @QueryParameter int maxIssuesFromJqlSearch,
```
rantoniuk, line 1377: I _think_ this might not be needed and it should be handled automatically by Jelly [validation as described here](https://github.com/jenkinsci/jenkins/blob/master/core/src/main/resources/lib/form/number.jelly#L62) but I haven't tested that, can you give it a try? That would avoid a lot of changes.

rantoniuk (changes requested):

src/main/java/hudson/plugins/jira/JiraSite.java
Hardikrathod01: It is not working, if I remove this parameter. 

on click of Validate Settings button, Success message is shown even if the number is bigger than 5000. 

grafik

src/main/java/hudson/plugins/jira/JiraSession.java
```diff
@@ -51,9 +49,12 @@ public class JiraSession {
 
     private final String jiraSiteName;
 
+    private final int maxIssuesFromJqlSearch;
```
rantoniuk, line 52: also `Integer` here

src/main/java/hudson/plugins/jira/JiraRestService.java
```diff
@@ -227,7 +227,7 @@ public List<String> getProjectsKeys() {
         return keys;
     }
 
-    public List<Issue> getIssuesFromJqlSearch(String jqlSearch, Integer maxResults) throws TimeoutException {
+    public List<Issue> getIssuesFromJqlSearch(String jqlSearch, int maxResults) throws TimeoutException {
```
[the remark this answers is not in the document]
Hardikrathod01: yes, I think that is why suddenly some tests started failing.

src/main/java/hudson/plugins/jira/JiraSite.java
```diff
@@ -646,6 +659,17 @@ public void setUpdateJiraIssueForAllStatus(boolean updateJiraIssueForAllStatus)
         this.updateJiraIssueForAllStatus = updateJiraIssueForAllStatus;
     }
 
+    @DataBoundSetter
+    public void setMaxIssuesFromJqlSearch(int maxIssuesFromJqlSearch) {
+        this.maxIssuesFromJqlSearch = maxIssuesFromJqlSearch > MAX_ALLOWED_ISSUES_FROM_JQL
+                ? MAX_ALLOWED_ISSUES_FROM_JQL
+                : maxIssuesFromJqlSearch;
+    }
```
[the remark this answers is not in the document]
Hardikrathod01: Just setting the default value is not enough. 
It still resulted in value 0 for `maxIssuesFromJqlSearch`  in the method `readResolve()`. And then in turn calling setter by passing 0 will obviously set the value to 0. So default became ineffective.

So I gave it a try with this approach. 

My test steps: 

1. Install plugin version 3.16 and add Jira site.
2. Update the plugin from the hpi file generated from source code. 
3. Restart of jenkins was needed for the update to take effect. 
4. Then the default value of 100 is set as per the below result

image

rantoniuk:
Great job, thanks for contribution!

PR #707, opened 2025-06-06
Add JIRA_ISSUES_SIZE to the JiraEnvironmentVariableBuilder

rantoniuk:
Contributed by @mupakoz in https://github.com/jenkinsci/jira-plugin/pull/119

Adds JiraIssuesSizeEnvironmentVariableBuilder build step which enables saving number of issues matching a JQL query in a environment variable of specified name.

I came up with this idea for a new feature when I wanted to fail the build in case there are issues on JIRA that were resolved but still untested. Thanks to the plugin we can save the number of issues we are interested in and then use it in the next steps (for example a shell script or a conditional step).

### DIFF
--- a/src/main/java/hudson/plugins/jira/JiraEnvironmentContributingAction.java
+++ b/src/main/java/hudson/plugins/jira/JiraEnvironmentContributingAction.java
@@ -33,7 +33,7 @@ public class JiraEnvironmentContributingAction extends InvisibleAction implement
         return jiraUrl;
     }
 
-    public JiraEnvironmentContributingAction(String issuesList, int issuesSize, String jiraUrl) {
+    public JiraEnvironmentContributingAction(String issuesList, Integer issuesSize, String jiraUrl) {
         this.issuesList = issuesList;
         this.issuesSize = issuesSize;
         this.jiraUrl = jiraUrl;


src/main/java/hudson/plugins/jira/JiraEnvironmentVariableBuilder.java
```diff
@@ -21,10 +21,12 @@
 public class JiraEnvironmentVariableBuilder extends Builder {
 
     private AbstractIssueSelector issueSelector;
+    private final String issuesSizeVariableName;
 
     @DataBoundConstructor
-    public JiraEnvironmentVariableBuilder(AbstractIssueSelector issueSelector) {
+    public JiraEnvironmentVariableBuilder(AbstractIssueSelector issueSelector, String issuesSizeVariableName) {
```
rantoniuk: public constructor: backward compatibilty check. Builder pattern?

src/main/java/hudson/plugins/jira/JiraEnvironmentContributingAction.java
```diff
@@ -16,25 +16,41 @@ public class JiraEnvironmentContributingAction extends InvisibleAction implement
...
 
+    public Integer getNumberOfIssues() {
+        return issuesSize;
+    }
+
     public String getJiraUrl() {
         return jiraUrl;
     }
 
-    public JiraEnvironmentContributingAction(String issuesList, String jiraUrl) {
```
rantoniuk, line 29: public constructor: backward compatibilty check. Builder pattern?

Issue #675, opened 2018-12-19
[JENKINS-55263] show only released version parameters


rantoniuk:
<p>It would be awesome if "Referencing JIRA Release version " could show only released versions (right now it can show unreleased or unreleased AND released).</p>

---
<details><summary><i>Originally reported by <a title="Foxy BOA" href="https://issues.jenkins.io/secure/ViewProfile.jspa?name=foxyboa">foxyboa</a>, imported from: <a href="https://issues.jenkins.io/browse/JENKINS-55263" target="_blank">Jira plugin. Show only released version parameters</a></i></summary>
<i><ul>
<li><b>status</b>: Open
<li><b>priority</b>: Minor
<li><b>resolution</b>: Unresolved
<li><b>imported</b>: 2025-05-30
</ul></i>
</details>

Hardikrathod01:
I think this issue is asking for the similar enhancement: https://github.com/jenkinsci/jira-plugin/issues/330

So I think if I add a new checkbox to show unreleased versions, then it would also solve this. 
So in the end there will be one checkbox each for Released, Unreleased and Archived. 

If this sounds good to you, I think I know the approach how to implement it. 

In ParameterDefinition, filter logic can be modified to filter the issues based on each checkbox selected. If none is selected we show released and unreleased both.

rantoniuk:
> So in the end there will be one checkbox each for Released, Unreleased and Archived.

That sounds good, we just need to make sure it is backwards-compatible (i.e. that the behavior doesn't change when the new option is added).

PR #708, opened 2025-06-07
Issue 675 add option show only released versions

Hardikrathod01:
Adding a new parameter to show only released versions. 
This PR addresses the issue: https://github.com/jenkinsci/jira-plugin/issues/675

It is tested in local jenkins with jira cloud. 

### DIFF
--- a/src/main/java/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinition.java
+++ b/src/main/java/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinition.java
@@ -10,6 +10,7 @@ import hudson.plugins.jira.JiraSession;
 import hudson.plugins.jira.JiraSite;
 import java.io.IOException;
 import java.util.List;
+import java.util.Objects;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import net.sf.json.JSONObject;
@@ -24,6 +25,7 @@ public class JiraVersionParameterDefinition extends ParameterDefinition {
     private String projectKey;
     private boolean showReleased = false;
     private boolean showArchived = false;
+    private boolean showOnlyReleased = false;
     private Pattern pattern = null;
 
     @DataBoundConstructor
@@ -33,12 +35,14 @@ public class JiraVersionParameterDefinition extends ParameterDefinition {
             String jiraProjectKey,
             String jiraReleasePattern,
             String jiraShowReleased,
-            String jiraShowArchived) {
+            String jiraShowArchived,
+            String jiraShowOnlyReleased) {
         super(name, description);
         setJiraProjectKey(jiraProjectKey);
         setJiraReleasePattern(jiraReleasePattern);
         setJiraShowReleased(jiraShowReleased);
         setJiraShowArchived(jiraShowArchived);
+        setShowOnlyReleased(jiraShowOnlyReleased);
     }
 
     @Override
@@ -90,17 +94,31 @@ public class JiraVersionParameterDefinition extends ParameterDefinition {
             }
         }
 
-        // Filter released versions
-        if (!showReleased && version.isReleased()) {
-            return false;
+        boolean isReleased = version.isReleased();
+        boolean isArchived = version.isArchived();
+        boolean allOptionsSelected = showReleased && showOnlyReleased && showArchived;
+
+        if (allOptionsSelected) {
+            return true;
         }
 
-        // Filter archived versions
-        if (!showArchived && version.isArchived()) {
-            return false;
+        if (showReleased && !isArchived) {
+            return true;
         }
 
-        return true;
+        if (showArchived && isArchived) {
+            return true;
+        }
+
+        if (showOnlyReleased) {
+            return isReleased;
+        }
+
+        if (!showReleased && !showArchived && !showOnlyReleased && !isReleased && !isArchived) {
+            return true;
+        }
+
+        return false;
     }
 
     public String getJiraReleasePattern() {
@@ -142,6 +160,14 @@ public class JiraVersionParameterDefinition extends ParameterDefinition {
         this.showArchived = Boolean.parseBoolean(showArchived);
     }
 
+    public String getJiraShowOnlyReleased() {
+        return Boolean.toString(showOnlyReleased);
+    }
+
+    public void setShowOnlyReleased(String showOnlyReleased) {
+        this.showOnlyReleased = Boolean.parseBoolean(showOnlyReleased);
+    }
+
     @Extension
     @Symbol("jiraReleaseVersion")
     public static class DescriptorImpl extends ParameterDescriptor {
@@ -158,6 +184,13 @@ public class JiraVersionParameterDefinition extends ParameterDefinition {
         public Result(final Version version) {
             this.name = version.getName();
             this.id = version.getId();
+        }
+
+        @Override
+        public boolean equals(Object o) {
+            if (o == null || getClass() != o.getClass()) return false;
+            Result result = (Result) o;
+            return Objects.equals(name, result.name) && Objects.equals(id, result.id);
         }
     }
 }
--- a/src/main/resources/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinition/config.jelly
+++ b/src/main/resources/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinition/config.jelly
@@ -16,6 +16,9 @@
 	<f:entry title="${%Show Archived Versions}" field="jiraShowArchived">
         <f:checkbox />
     </f:entry>
+    <f:entry title="${%Show Only Released Versions}" field="jiraShowOnlyReleased">
+        <f:checkbox />
+    </f:entry>
 	<f:entry title="${%Release Name Pattern}" field="jiraReleasePattern">
         <f:textbox />
     </f:entry>
--- a/src/test/java/hudson/plugins/jira/versionparameter/JiraVersionParameterFilterTest.java
+++ b/src/test/java/hudson/plugins/jira/versionparameter/JiraVersionParameterFilterTest.java
@@ -63,9 +63,7 @@ public class JiraVersionParameterFilterTest {
             } catch (IOException e) {
                 throw new RuntimeException(e);
             }
-            List<JiraVersionParameterDefinition.Result> expected = new ArrayList<>();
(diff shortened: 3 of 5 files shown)


src/main/java/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinition.java
```diff
@@ -90,17 +94,31 @@ private boolean match(Version version) {
             }
         }
 
-        // Filter released versions
-        if (!showReleased && version.isReleased()) {
-            return false;
+        boolean isReleased = version.isReleased();
+        boolean isArchived = version.isArchived();
+        boolean allOptionsSelected = showReleased && showOnlyReleased && showArchived;
```
rantoniuk: Adjust the tests to cover the line fully

src/main/java/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinition.java
```diff
@@ -90,17 +94,31 @@ private boolean match(Version version) {
...
+        }
+
+        if (showOnlyReleased) {
+            return isReleased;
         }
 
-        // Filter archived versions
-        if (!showArchived && version.isArchived()) {
-            return false;
+        if (!showReleased && !showArchived && !showOnlyReleased && !isReleased && !isArchived) {
```
rantoniuk: Adjust tests to cover this line fully

src/test/java/hudson/plugins/jira/versionparameter/JiraVersionParameterFilterTest.java
```diff
@@ -0,0 +1,154 @@
+package hudson.plugins.jira.versionparameter;
```
rantoniuk: Why are these tests in a separate file? If there's no specific need, move them to `JiraVersionParameterDefinitionTest.java`.

src/test/java/hudson/plugins/jira/versionparameter/JiraVersionParameterFilterTest.java
```diff
@@ -0,0 +1,154 @@
...
+        jenkins = mock(Jenkins.class);
+        globalConfig = mock(JiraGlobalConfiguration.class);
+
+        when(request2.findAncestorObject(Job.class)).thenReturn(job);
+        when(site.getSession(job)).thenReturn(session);
+        when(session.getVersions("PROJ")).thenReturn(versions);
+    }
+
+    @Test
+    void showReleaseTest() {
```
rantoniuk: Test methods do not have to have `Test` suffix.

src/test/java/hudson/plugins/jira/versionparameter/JiraVersionParameterFilterTest.java
```diff
@@ -0,0 +1,154 @@
...
+    private StaplerRequest2 request2;
+    Jenkins jenkins;
+    private JiraGlobalConfiguration globalConfig;
+
+    @BeforeEach
+    void createMocksAndVersions() {
+        versions.add(new ExtendedVersion(null, null, "1.0", "", false, true, null, null));
+        versions.add(new ExtendedVersion(null, null, "1.1", "", false, false, null, null));
+        versions.add(new ExtendedVersion(null, null, "1.2", "", true, false, null, null));
+        r1 = new JiraVersionParameterDefinition.Result(new Version(null, null, "1.0", "", false, true, null));
```
rantoniuk: Rename those variables so that they are actually meaningful.

src/test/java/hudson/plugins/jira/versionparameter/JiraVersionParameterFilterTest.java
```diff
@@ -0,0 +1,154 @@
...
+
+    List<ExtendedVersion> versions = new ArrayList<>();
+    JiraVersionParameterDefinition.Result r1;
+    JiraVersionParameterDefinition.Result r2;
+    JiraVersionParameterDefinition.Result r3;
+    private Job<?, ?> job;
+    private JiraSite site;
+    private JiraSession session;
+    private StaplerRequest2 request2;
+    Jenkins jenkins;
```
rantoniuk: See also `@WithoutJenkins`

src/test/java/hudson/plugins/jira/versionparameter/JiraVersionParameterFilterTest.java
```diff
@@ -0,0 +1,154 @@
...
+    private JiraGlobalConfiguration globalConfig;
+
+    @BeforeEach
+    void createMocksAndVersions() {
+        versions.add(new ExtendedVersion(null, null, "1.0", "", false, true, null, null));
+        versions.add(new ExtendedVersion(null, null, "1.1", "", false, false, null, null));
+        versions.add(new ExtendedVersion(null, null, "1.2", "", true, false, null, null));
```
rantoniuk: Missing one more: released and archived

src/test/java/hudson/plugins/jira/versionparameter/JiraVersionParameterFilterTest.java
```diff
@@ -0,0 +1,154 @@
...
+    private Job<?, ?> job;
+    private JiraSite site;
+    private JiraSession session;
+    private StaplerRequest2 request2;
+    Jenkins jenkins;
+    private JiraGlobalConfiguration globalConfig;
+
+    @BeforeEach
+    void createMocksAndVersions() {
+        versions.add(new ExtendedVersion(null, null, "1.0", "", false, true, null, null));
```
rantoniuk: ExtendedVersion inherits from Version, so you should be able to re-use them in Result creation below

src/main/java/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinition.java
```diff
@@ -33,12 +35,14 @@ public JiraVersionParameterDefinition(
             String jiraProjectKey,
             String jiraReleasePattern,
             String jiraShowReleased,
-            String jiraShowArchived) {
+            String jiraShowArchived,
+            String jiraShowOnlyReleased) {
```
rantoniuk: Implement and align the same logic as in Jira, i.e.:
- released
- unreleased
- archived.

That will fullfil the issue in question and the logic should be simpler.

Screenshot 2025-06-09 at 12 01 33 AM

src/main/java/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinition.java
```diff
@@ -159,5 +185,12 @@ public Result(final Version version) {
             this.name = version.getName();
             this.id = version.getId();
         }
+
+        @Override
+        public boolean equals(Object o) {
+            if (o == null || getClass() != o.getClass()) return false;
+            Result result = (Result) o;
+            return Objects.equals(name, result.name) && Objects.equals(id, result.id);
```
rantoniuk, line 189: Adjust tests, this is important method that needs to be tested (if needed at all)

rantoniuk (changes requested):

src/test/java/hudson/plugins/jira/versionparameter/JiraVersionParameterFilterTest.java
Hardikrathod01: I did not add this specific case, because Jira treats released and archived as archived in the frontend. 

In the below SS, API returns version 1.0 as released and archived
image

but, frontend shows that version only when archived is selected.
image

if I select released as the filter, it does not show version 1.0.

src/main/java/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinition.java
Hardikrathod01: I am not sure why is it still saying that 189 is not fully covered. 
When I run the test with coverage in intellij, it shows this method is fully covered.

src/main/java/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinition.java
```diff
@@ -142,6 +156,14 @@ public void setJiraShowArchived(String showArchived) {
         this.showArchived = Boolean.parseBoolean(showArchived);
     }
 
+    public String getJiraShowUnreleased() {
+        return Boolean.toString(showUnreleased);
```
rantoniuk, line 160: That's easy to cover

src/main/java/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinition.java
```diff
@@ -159,5 +181,12 @@ public Result(final Version version) {
             this.name = version.getName();
             this.id = version.getId();
         }
+
+        @Override
+        public boolean equals(Object o) {
```
rantoniuk, line 186: API Note:

It is generally necessary to override the [hashCode](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) method whenever this method is overridden, so as to maintain the general contract for the hashCode method, which states that equal objects must have equal hash codes.

src/main/java/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinition.java
rantoniuk: I think the missing branch might be comparing comparing objects with the same name but different IDs (or the other way round) - I don't see this test case.

src/main/resources/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinition/config.jelly
```diff
@@ -12,6 +12,9 @@
     </f:entry>
     <f:entry title="${%Show Released Versions}" field="jiraShowReleased">
```
rantoniuk, line 13: I think it would be good to add a help text saying that "By default when all checkboxes are unchecked, all versions are displayed" - see [inline-help](https://www.jenkins.io/doc/developer/forms/jelly-form-controls/#add-an-inline-help-to-a-form-element)

rantoniuk (changes requested):
overally almost ready to merge, small changes required

Issue #702, opened 2025-05-30
Log an appropriate AccessDenied message instead of throwing a RestClientException stacktrace


rantoniuk:
When the plugin finds a pattern that matches the configured patterns, it tries to access the issue status from the Jira instance. If it does not have the required permissions, it throws an ugly stacktrace instead of just stating in the build log that it doesn't have access to that issue.

This might be the beginning (or part of) a larger re-factoring of logging so that:
- Rest exceptions are not thrown back (at least if DEBUG is not enabled) and instead a user friendly message is logged
- that the log message goes to the build log and not (only) the Jenkins log, so that it's visible to the actual user accessing the build logs via Jenkins UI.

> ```
> Jira REST client get issue error. cause: RestClientException{statusCode=Optional.of(403), errorCollections=[ErrorCollection{status=403, errors={}, errorMessages=[You do not have the permission to see the specified issue.]}]}
> RestClientException{statusCode=Optional.of(403), errorCollections=[ErrorCollection{status=403, errors={}, errorMessages=[You do not have the permission to see the specified issue.]}]}
> 	at PluginClassLoader for jira//com.atlassian.jira.rest.client.internal.async.AbstractAsynchronousRestClient$2.apply(AbstractAsynchronousRestClient.java:189)
> 	at PluginClassLoader for jira//com.atlassian.jira.rest.client.internal.async.AbstractAsynchronousRestClient$2.apply(AbstractAsynchronousRestClient.java:183)
> 	at PluginClassLoader for jira//com.atlassian.httpclient.api.ResponsePromiseMapFunction.apply(ResponsePromiseMapFunction.java:49)
> 	at PluginClassLoader for jira//com.atlassian.httpclient.api.ResponsePromiseMapFunction.apply(ResponsePromiseMapFunction.java:10)
> 	at PluginClassLoader for jira//io.atlassian.util.concurrent.Promises$OfStage.lambda$fold$4(Promises.java:332)
> 	at PluginClassLoader for jira//io.atlassian.util.concurrent.Promises.lambda$biFunction$7(Promises.java:422)
> 	at java.base/java.util.concurrent.CompletableFuture.uniHandle(CompletableFuture.java:934)
> 	at java.base/java.util.concurrent.CompletableFuture$UniHandle.tryFire(CompletableFuture.java:911)
> 	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510)
[...]

PR #711, opened 2025-06-17
log appropriate error messages to build console

Hardikrathod01:

### Related issues

https://github.com/jenkinsci/jira-plugin/issues/702
### Changes
This PR implements error handling for the methods of JiraRestService class. I have added a new class JiraException and every method from JiraRestService can throw JiraException and calling methods can log the related messages. 

### Tests

 

- [x] I have tested my changes with a Jira Cloud

### Checklist

- [x] Make sure you are requesting the pull request from a _custom branch_ and **not your master/main branch**
- [ ] Apply relevant PR labels, at least one of `feature` or `bugfix` or `documentation`

### DIFF
--- a/src/main/java/hudson/plugins/jira/JiraException.java
+++ b/src/main/java/hudson/plugins/jira/JiraException.java
@@ -1,0 +1,11 @@
+package hudson.plugins.jira;
+
+public class JiraException extends Exception {
+    public JiraException(String message, Throwable cause) {
+        super(message, cause);
+    }
+
+    public JiraException(String message) {
+        super(message);
+    }
+}
--- a/src/main/java/hudson/plugins/jira/JiraRestService.java
+++ b/src/main/java/hudson/plugins/jira/JiraRestService.java
@@ -247,7 +247,7 @@ public class JiraRestService {
         }
     }
 
-    public List<ExtendedVersion> getVersions(String projectKey) {
+    public List<ExtendedVersion> getVersions(String projectKey) throws JiraException {
         final URIBuilder builder =
                 new URIBuilder(uri).setPath(String.format("%s/project/%s/versions", baseApiPath, projectKey));
 
@@ -259,6 +259,7 @@ public class JiraRestService {
             decoded = objectMapper.readValue(content.asString(), new TypeReference<List<Map<String, Object>>>() {});
         } catch (Exception e) {
             LOGGER.log(WARNING, "Jira REST client get versions error. cause: " + e.getMessage(), e);
+            throw new JiraException("[Jira Error] Jira REST client get versions error. cause: " + e.getMessage(), e);
         }
 
         return decoded.stream()
--- a/src/main/java/hudson/plugins/jira/JiraSession.java
+++ b/src/main/java/hudson/plugins/jira/JiraSession.java
@@ -142,7 +142,7 @@ public class JiraSession {
      * @param projectKey The key for the project
      * @return An array of versions
      */
-    public List<ExtendedVersion> getVersions(String projectKey) {
+    public List<ExtendedVersion> getVersions(String projectKey) throws JiraException {
         LOGGER.fine("Fetching versions from project: " + projectKey);
         return service.getVersions(projectKey);
     }
@@ -154,7 +154,7 @@ public class JiraSession {
      * @param name       The version name
      * @return A RemoteVersion, or null if not found
      */
-    public ExtendedVersion getVersionByName(String projectKey, String name) {
+    public ExtendedVersion getVersionByName(String projectKey, String name) throws JiraException {
         LOGGER.fine("Fetching versions from project: " + projectKey);
         List<ExtendedVersion> versions = getVersions(projectKey);
         if (versions == null) {
@@ -219,7 +219,7 @@ public class JiraSession {
      * @param version    The replacement version
      * @param query      The JQL Query
      */
-    public void migrateIssuesToFixVersion(String projectKey, String version, String query) throws TimeoutException {
+    public void migrateIssuesToFixVersion(String projectKey, String version, String query) throws TimeoutException, JiraException {
 
         Version newVersion = getVersionByName(projectKey, version);
         if (newVersion == null) {
@@ -249,7 +249,7 @@ public class JiraSession {
      * @param query       The JQL Query
      */
     public void replaceFixVersion(String projectKey, String fromVersion, String toVersion, String query)
-            throws TimeoutException {
+            throws TimeoutException, JiraException {
 
         Version newVersion = getVersionByName(projectKey, toVersion);
         if (newVersion == null) {
@@ -305,7 +305,7 @@ public class JiraSession {
      * @param version    The version to add
      * @param query      The JQL Query
      */
-    public void addFixVersion(String projectKey, String version, String query) throws TimeoutException {
+    public void addFixVersion(String projectKey, String version, String query) throws TimeoutException, JiraException {
 
         Version newVersion = getVersionByName(projectKey, version);
         if (newVersion == null) {
--- a/src/main/java/hudson/plugins/jira/JiraSession.java
+++ b/src/main/java/hudson/plugins/jira/JiraSession.java
@@ -219,7 +219,8 @@ public class JiraSession {
      * @param version    The replacement version
      * @param query      The JQL Query
      */
-    public void migrateIssuesToFixVersion(String projectKey, String version, String query) throws TimeoutException, JiraException {
+    public void migrateIssuesToFixVersion(String projectKey, String version, String query)
+            throws TimeoutException, JiraException {
 
         Version newVersion = getVersionByName(projectKey, version);
         if (newVersion == null) {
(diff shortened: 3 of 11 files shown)


src/main/java/hudson/plugins/jira/JiraException.java
```diff
@@ -0,0 +1,8 @@
+package hudson.plugins.jira;
```
rantoniuk: That class doesn't provide any additional value, why not use `com.atlassian.jira.rest.client.api.RestClientException` directly?

Hardikrathod01: I wasn’t sure if I would need anything else when I implement it for other methods too. But I will remove it if nothing else is required.

src/main/java/hudson/plugins/jira/JiraIssueUpdateBuilder.java
```diff
@@ -108,6 +105,8 @@ public void perform(Run<?, ?> run, EnvVars env, TaskListener listener) throws In
             listener.getLogger().println(Messages.JiraIssueUpdateBuilder_Failed());
             e.printStackTrace(listener.getLogger());
             run.setResult(Result.FAILURE);
+        } catch (RestClientException re) {
+            listener.getLogger().println(re.getMessage());
```
rantoniuk: Why would we treat `TimeoutException` differently from `RestClientException` ?

src/main/java/hudson/plugins/jira/JiraJobAction.java
```diff
@@ -129,6 +130,8 @@ public void onStarted(WorkflowRun workflowRun, TaskListener listener) {
                     setAction(parent, site);
                 } catch (IOException e) {
                     LOGGER.log(Level.WARNING, "Could not set JiraJobAction for <" + parent.getFullName() + ">", e);
+                } catch (RestClientException e) {
```
rantoniuk: Same here, why treat those differently?

src/main/java/hudson/plugins/jira/JiraRestService.java
```diff
@@ -174,10 +176,12 @@ public Issue getIssue(String issueKey) {
                     && ((RestClientException) e.getCause()).getStatusCode().isPresent()
                     && ((RestClientException) e.getCause()).getStatusCode().get() == 404) {
                 LOGGER.log(INFO, "Issue '" + issueKey + "' not found in Jira.");
+                throw new RestClientException("[Jira Error] Issue '" + issueKey + "' not found in Jira.", e.getCause());
```
rantoniuk:
```suggestion
                throw new RestClientException("[Jira] Issue '" + issueKey + "' not found in Jira.", e.getCause());
```

src/main/java/hudson/plugins/jira/JiraRestService.java
```diff
@@ -174,10 +176,12 @@ public Issue getIssue(String issueKey) {
...
             } else {
                 LOGGER.log(WARNING, "Jira REST client get issue error. cause: " + e.getMessage(), e);
+                throw new RestClientException(
+                        "[Jira Error] Jira REST client get issue error. cause: " + e.getMessage(), e.getCause());
```
rantoniuk:
```suggestion
                        "[Jira] Jira REST client get issue error. cause: " + e.getMessage(), e.getCause());
```

src/main/java/hudson/plugins/jira/JiraRestService.java
```diff
@@ -193,7 +197,8 @@ public List<IssueType> getIssueTypes() {
                     .collect(Collectors.toList());
         } catch (Exception e) {
             LOGGER.log(WARNING, "Jira REST client get issue types error. cause: " + e.getMessage(), e);
-            return Collections.emptyList();
+            throw new RestClientException(
+                    "[Jira Error] Jira REST client get issue types error. cause: " + e.getMessage(), e.getCause());
```
rantoniuk:
```suggestion
                    "[Jira] Jira REST client get issue types error. cause: " + e.getMessage(), e.getCause());
```

src/main/java/hudson/plugins/jira/JiraRestService.java
```diff
@@ -209,7 +214,8 @@ public List<Priority> getPriorities() {
                     .collect(Collectors.toList());
         } catch (Exception e) {
             LOGGER.log(WARNING, "Jira REST client get priorities error. cause: " + e.getMessage(), e);
-            return Collections.emptyList();
+            throw new RestClientException(
+                    "[Jira Error] Jira REST client get priorities error. cause: " + e.getMessage(), e.getCause());
```
rantoniuk:
```suggestion
                    "[Jira] Jira REST client get priorities error. cause: " + e.getMessage(), e.getCause());
```

src/main/java/hudson/plugins/jira/JiraRestService.java
```diff
@@ -209,7 +214,8 @@ public List<Priority> getPriorities() {
                     .collect(Collectors.toList());
         } catch (Exception e) {
```
rantoniuk: Catch `RestClientException` here

src/main/java/hudson/plugins/jira/JiraRestService.java
```diff
@@ -219,6 +225,8 @@ public List<String> getProjectsKeys() {
             projects = jiraRestClient.getProjectClient().getAllProjects().get(timeout, TimeUnit.SECONDS);
         } catch (Exception e) {
```
rantoniuk: Catch `RestClientException` here

src/main/java/hudson/plugins/jira/JiraRestService.java
```diff
@@ -219,6 +225,8 @@ public List<String> getProjectsKeys() {
...
             LOGGER.log(WARNING, "Jira REST client get project keys error. cause: " + e.getMessage(), e);
+            throw new RestClientException(
+                    "[Jira Error] Jira REST client get project keys error. cause: " + e.getMessage(), e.getCause());
```
rantoniuk:
```suggestion
                    "[Jira] Jira REST client get project keys error. cause: " + e.getMessage(), e.getCause());
```

src/main/java/hudson/plugins/jira/JiraRestService.java
```diff
@@ -243,22 +252,26 @@ public List<Issue> getIssuesFromJqlSearch(String jqlSearch, Integer maxResults)
...
+                    "[Jira Error] Jira REST client get issue from jql search error. cause: " + e.getMessage(),
+                    e.getCause());
         }
     }
 
     public List<ExtendedVersion> getVersions(String projectKey) {
         final URIBuilder builder =
                 new URIBuilder(uri).setPath(String.format("%s/project/%s/versions", baseApiPath, projectKey));
 
-        List<Map<String, Object>> decoded = Collections.emptyList();
```
rantoniuk: No need to touch that I guess

rantoniuk (changes requested):
Sorry for long waiting time for this review. Before we move further, some basic changes are needed around how the exceptions are handled.

In the areas where you modified the code: 
- replace the use of generic `Exception e` with the specific `RestClientException` so that this PR makes actual sense in cleaning up the exception handling logic
- replace `[Jira error]` with `[Jira]`

src/main/java/hudson/plugins/jira/JiraIssueUpdateBuilder.java
Hardikrathod01: I had kept `TimeoutException` because the method `getIssuesFromJqlSearch` in `JiraRestService` was throwing it separately. And the rest three exceptions from the method namely, `CancellationException`, `ExecutionException`, and `InterruptedException` were handled and rethrown as a `RestClientException`. 
But I think, all of them can be wrapped in `RestClientException`.

src/main/java/hudson/plugins/jira/JiraJobAction.java
Hardikrathod01: Here the `IOException` is being thrown from multiple places which I think is not specific to JiraRestClient.
For example: `job.save();` throws IOException. 
Another place from where it is thrown is: `String decodedJobName = URLDecoder.decode(job.getName(), "UTF-8");`

So I could handle those IOExceptions and rethrow as RestClientException.

rantoniuk: Why not `IOException e | RestClientException e` ?

src/main/java/hudson/plugins/jira/JiraCreateIssueNotifier.java
```diff
@@ -2,6 +2,7 @@
 
 import static hudson.plugins.jira.JiraRestService.BUG_ISSUE_TYPE_ID;
 
+import com.atlassian.jira.rest.client.api.RestClientException;
```
rantoniuk, line 5: I think we're going the wrong way here - this is introducing throwing exception to this class which is complicating it more rather than simplyfing. 

We already have a split between JiraSession and JiraRestService, so the domain classes should not any exceptions - instead they should be thrown (or logged) in JiraSession or JiraRestService.

What is your view on this?

Hardikrathod01: Comparatively to me the current approach looked simpler. 
If we want to log it in JiraSession or JiraRestService, then we need to pass on listener's logger somehow. 
Which we might have to do at 5-6 places from wherever the methods are called. 

Or second option is to set the logger as a field in JiraRestService while calling it's method. That would simplify logging further.

rantoniuk:
@Hardikrathod01 sorry for picking this up so late, could you resolve the conflicts please?

PR #713, opened 2025-06-19
Remove workspace requirement for SimpleBuildSteps

res0nance:

### Related issues

https://github.com/jenkinsci/jenkins/pull/4829/
[JENKINS-46175](https://issues.jenkins.io/browse/JENKINS-46175)

### Changes

Remove requirement for a workspace in all SimpleBuildSteps

We do not use the workspace at all in any of our steps so this allows us to remove the requirement. This also allows the steps to be used outside of a `node` block

### Tests

 

### Checklist

- [x] Make sure you are requesting the pull request from a _custom branch_ and **not your master/main branch**
- [x] Apply relevant PR labels, at least one of `feature` or `bugfix` or `documentation`

### DIFF
--- a/src/main/java/hudson/plugins/jira/JiraIssueUpdater.java
+++ b/src/main/java/hudson/plugins/jira/JiraIssueUpdater.java
@@ -1,7 +1,7 @@
 package hudson.plugins.jira;
 
+import hudson.EnvVars;
 import hudson.Extension;
-import hudson.FilePath;
 import hudson.Launcher;
 import hudson.matrix.MatrixAggregatable;
 import hudson.matrix.MatrixAggregator;
@@ -52,8 +52,7 @@ public class JiraIssueUpdater extends Recorder implements MatrixAggregatable, Si
     }
 
     @Override
-    public void perform(Run<?, ?> run, FilePath workspace, Launcher launcher, TaskListener listener)
-            throws InterruptedException, IOException {
+    public void perform(Run<?, ?> run, EnvVars env, TaskListener listener) throws InterruptedException, IOException {
         // Don't do anything for individual matrix runs.
         if (run instanceof MatrixRun) {
             return;
@@ -68,6 +67,11 @@ public class JiraIssueUpdater extends Recorder implements MatrixAggregatable, Si
             throw new IllegalArgumentException(
                     "Unsupported run type " + run.getClass().getName());
         }
+    }
+
+    @Override
+    public boolean requiresWorkspace() {
+        return false;
     }
 
     @Override
--- a/src/test/java/hudson/plugins/jira/JiraIssueUpdateBuilderTest.java
+++ b/src/test/java/hudson/plugins/jira/JiraIssueUpdateBuilderTest.java
@@ -20,9 +20,15 @@ import hudson.model.Result;
 import hudson.model.TaskListener;
 import java.io.IOException;
 import java.io.PrintStream;
+import java.util.Collections;
 import java.util.concurrent.TimeoutException;
+import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
+import org.jenkinsci.plugins.workflow.job.WorkflowJob;
+import org.jenkinsci.plugins.workflow.job.WorkflowRun;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.jvnet.hudson.test.JenkinsRule;
+import org.jvnet.hudson.test.junit.jupiter.WithJenkins;
 
 class JiraIssueUpdateBuilderTest {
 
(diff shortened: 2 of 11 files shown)


rantoniuk:
Please fill in the PR description accordingly and describe what is this PR about / why do you need this change.

res0nance:
> Please fill in the PR description accordingly and describe what is this PR about / why do you need this change.

Done I have filled up the description with larger context on the Jenkins ecosystem

src/main/java/hudson/plugins/jira/JiraIssueUpdater.java
```diff
@@ -1,7 +1,7 @@
 package hudson.plugins.jira;
 
+import hudson.EnvVars;
 import hudson.Extension;
-import hudson.FilePath;
 import hudson.Launcher;
```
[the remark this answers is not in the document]
res0nance: This is a mistake used in createAggregator

rantoniuk:
Can you share a simple pipeline definition that you tested this change with? I have one scenario in mind that I'd like to verify (since no tests were changed).

res0nance:
> Can you share a simple pipeline definition that you tested this change with? I have one scenario in mind that I'd like to verify (since no tests were changed).

Something like https://stackoverflow.com/questions/35213266/updating-jira-tickets-from-jenkins-workflow-jenkinsfile should work. I've not been able to test this with Jira as I don't have a Jira project i can use.

But the same step should work just not in in a `node` block it should also work in a `node` block as well

rantoniuk:
> But the same step should work just not in in a node block it should also work in a node block as well

I agree. Since you were not able to test this in practice, would you be able to extend [the test cases](https://github.com/jenkinsci/jira-plugin/tree/master/src/test/java/hudson/plugins/jira/pipeline) for this change?

By the way, [there is a Jira Cloud instance](https://github.com/jenkinsci/jira-plugin?tab=readme-ov-file#tests) that you can also use for testing - let me know if you need access.

res0nance:
> > But the same step should work just not in in a node block it should also work in a node block as well
> 
> I agree. Since you were not able to test this in practice, would you be able to extend [the test cases](https://github.com/jenkinsci/jira-plugin/tree/master/src/test/java/hudson/plugins/jira/pipeline) for this change?
> 
> By the way, [there is a Jira Cloud instance](https://github.com/jenkinsci/jira-plugin?tab=readme-ov-file#tests) that you can also use for testing - let me know if you need access.

I've extended the tests where I can and I think this is reasonable as is.

src/test/java/hudson/plugins/jira/JiraIssueUpdateBuilderTest.java
```diff
@@ -37,8 +44,11 @@ class JiraIssueUpdateBuilderTest {
     private Result result;
     private JiraSite site;
 
+    private JenkinsRule r;
+
     @BeforeEach
-    void createMocks() throws IOException, InterruptedException {
+    void createMocks(JenkinsRule r) throws IOException, InterruptedException {
+        this.r = r;
```
rantoniuk: that's not needed in @BeforeEach

src/test/java/hudson/plugins/jira/JiraIssueUpdateBuilderTest.java
```diff
@@ -96,4 +106,19 @@ void performProgressOK() throws InterruptedException, IOException, TimeoutExcept
         builder.perform(build, workspace, launcher, listener);
         assertThat(result, is(Result.SUCCESS));
     }
+
+    @Test
+    void testPipelineWithJiraSite() throws Exception {
```
rantoniuk: add the JenkinsRule in this method directly instead - see [JavaDocs](https://javadoc.jenkins.io/component/jenkins-test-harness/org/jvnet/hudson/test/junit/jupiter/WithJenkins.html)

src/test/java/hudson/plugins/jira/pipeline/IssueFieldUpdateStepTest.java
```diff
@@ -68,8 +68,11 @@ class IssueFieldUpdateStepTest {
     @Mock(strictness = Mock.Strictness.LENIENT)
     Job job;
 
+    private JenkinsRule r;
```
rantoniuk, line 71: same as in the other file, add parameter to `testPipeline()` method instead

src/main/java/hudson/plugins/jira/JiraIssueUpdater.java
```diff
@@ -70,6 +69,11 @@ public void perform(Run<?, ?> run, FilePath workspace, Launcher launcher, TaskLi
         }
     }
 
+    @Override
+    public boolean requiresWorkspace() {
+        return false;
```
rantoniuk, line 74: You covered all other classes apart from this one, can you try to cover this one as well?

rantoniuk (changes requested):
almost ready, small tweaks needed

src/test/java/hudson/plugins/jira/pipeline/IssueFieldUpdateStepTest.java
res0nance: We cannot because this class uses the Jcasc rule and some of the other tests rely on it

src/main/java/hudson/plugins/jira/JiraIssueUpdater.java
res0nance: Tried to get something reasonable in, but I couldn't get the SCM working. But it does test that filepaths are not required which is the change introduced in this PR

PR #746, opened 2025-10-11
feat(selector): select issues since last successful build (#453)

### LINKED ISSUE #453: Jira issue selector for all changes since last successful build
I need to select Jira issues for all changes since last succesful build.
DefaultSelector is only selecting from the current build.
Is there a way to use defaultselector maybe on a text file where I can aggregate all the changes since last successful build?

ahmedanwar1:
### Related issue

Resolves [#453](https://github.com/jenkinsci/jira-plugin/issues/453)

### Changes

- Added `SinceLastSuccessfulBuildIssueSelector` to collect JIRA issues from all builds since the last successful build, excluding the last successful build itself.
- Refactored `AbstractIssueSelector` and `DefaultIssueSelector` for reusability and consistent issue extraction logic across selectors.
- Added comprehensive unit tests for the new selector to ensure correct issue collection.

### Tests

 

- Verified that issues are correctly collected from multiple builds after the last successful build.
- Confirmed that issues from the last successful build are excluded.
- Checked that log messages reflect the correct number of builds and issues processed.

<br/>

### DIFF
--- a/src/main/resources/hudson/plugins/jira/Messages.properties
+++ b/src/main/resources/hudson/plugins/jira/Messages.properties
@@ -47,3 +47,4 @@ ErrorCommentingIssues=[Jira] Could not comment on some issues: {0}
 JiraSite.threadExecutorMinimunSize = Thread Executor Size must be at least {0} (higher values are recommended)
 JiraSite.timeoutMinimunValue = Connection timeout must be at least {0}
 JiraSite.readTimeoutMinimunValue = Read timeout must be at least {0}
+SinceLastSuccessfulBuildIssueSelector.DisplayName = Since last successful build selector
--- a/src/main/resources/hudson/plugins/jira/Messages.properties
+++ b/src/main/resources/hudson/plugins/jira/Messages.properties
@@ -47,4 +47,4 @@ ErrorCommentingIssues=[Jira] Could not comment on some issues: {0}
 JiraSite.threadExecutorMinimunSize = Thread Executor Size must be at least {0} (higher values are recommended)
 JiraSite.timeoutMinimunValue = Connection timeout must be at least {0}
 JiraSite.readTimeoutMinimunValue = Read timeout must be at least {0}
-SinceLastSuccessfulBuildIssueSelector.DisplayName = Since last successful build selector
+SinceLastSuccessfulBuildIssueSelector.DisplayName = Select issues since the last successful build
--- a/src/main/resources/hudson/plugins/jira/Messages.properties
+++ b/src/main/resources/hudson/plugins/jira/Messages.properties
@@ -50,3 +50,4 @@ JiraSite.timeoutMinimunValue = Connection timeout must be at least {0}
 JiraSite.readTimeoutMinimunValue = Read timeout must be at least {0}
 JiraIssueParameterDefinition.NoIssueMatchedSearch = No issues matched the search
 JiraVersionParameterDefinition.NoVersionsMatchedSearch = No version matched the search
+SinceLastSuccessfulBuildIssueSelector.DisplayName = Since last successful build selector
--- a/docs/features.md
+++ b/docs/features.md
@@ -468,6 +468,7 @@ Several steps above take an `issueSelector:` argument that decides *which* issue
 - **`JqlSelector('project = EX AND status = Open')`** — runs a JQL query.
 - **`P4Selector()`** — reads Perforce job IDs from the changelog instead of scanning for issue-key
   patterns. Requires the P4 plugin to be also installed.
+- **`SinceLastSuccessfulBuildSelector()`** — collects issues from the current build and all builds since the last successful build (excluding the last successful build itself). If no previous successful build exists, only the current build is scanned.
 
 ## What is not yet supported in Pipeline
 
--- a/src/main/java/hudson/plugins/jira/selector/AbstractIssueSelector.java
+++ b/src/main/java/hudson/plugins/jira/selector/AbstractIssueSelector.java
@@ -12,12 +12,10 @@ import hudson.plugins.jira.RunScmChangeExtractor;
 import hudson.plugins.jira.listissuesparameter.JiraIssueParameterValue;
 import hudson.scm.ChangeLogSet;
 import hudson.scm.ChangeLogSet.Entry;
-
 import java.util.Set;
 import java.util.logging.Logger;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-
 import org.apache.commons.lang.StringUtils;
 
 /**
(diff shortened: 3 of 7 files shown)


src/main/resources/hudson/plugins/jira/Messages.properties
```diff
@@ -47,3 +47,4 @@ ErrorCommentingIssues=[Jira] Could not comment on some issues: {0}
 JiraSite.threadExecutorMinimunSize = Thread Executor Size must be at least {0} (higher values are recommended)
 JiraSite.timeoutMinimunValue = Connection timeout must be at least {0}
 JiraSite.readTimeoutMinimunValue = Read timeout must be at least {0}
+SinceLastSuccessfulBuildIssueSelector.DisplayName = Since last successful build selector
```
rantoniuk, line 53:
```suggestion
SinceLastSuccessfulBuildIssueSelector.DisplayName = Select issues since the last successful build
```

rantoniuk (changes requested):
@ahmedanwar1 thanks for contributing! 

Before going into details:
- can you remove the `AbstractIssueSelector` changes? This class is Abstract on purpose and should not contain any logic and I don't why would you need those changes for the new selector you're providing

- please extend the documentation, i.e. [features.md](https://github.com/jenkinsci/jira-plugin/blob/master/docs/features.md) and preferably as well in the [use-cases.md](https://github.com/jenkinsci/jira-plugin/blob/master/docs/usage-examples.md), since this selector is a more complex one. If you wish, you can extract all selectors documetantion into a separate Markdown file.

ahmedanwar1:
Thanks for the feedback @rantoniuk 

I modified `AbstractIssueSelector` to avoid code duplication, both `DefaultIssueSelector` and the new selector need the same logic for extracting issues from changelogs and parameters. Having `SinceLastSuccessfulBuildIssueSelector` depend on `DefaultIssueSelector` methods seemed architecturally wrong since they're independent strategies

Would you prefer a separate utility class like `IssueSelectorUtils` for shared logic or is some duplication acceptable to keep selectors self contained or you suggest another approach?

and should this selector also collect issues from the build parameters like `DefaultIssueSelector` does or is changelog only sufficient?

rantoniuk:
> Would you prefer a separate utility class like IssueSelectorUtils for shared logic or is some duplication acceptable to keep selectors self contained or you suggest another approach?

I'm fine with both solutions, i.e. extend _DefaultIssueSelector_ or extract to _IssueSelectorUtils_.

> and should this selector also collect issues from the build parameters like DefaultIssueSelector does or is changelog only sufficient?

The reason _DefaultIssueSelector_ does this is described in [JENKINS-12312](https://issues.jenkins.io/browse/JENKINS-12312), so I think we need the same behavior here.

ahmedanwar1:
I may tend to prefer extracting the shared logic into an `IssueSelectorUtils` class. i think extending `DefaultIssueSelector` feels semantically off and would introduce tighter coupling between the selectors

ahmedanwar1:
Actually after looking deeper through the codebase, I realized `IssueSelectorUtils` won't work well. I found that `JobIssueSelector`  extends `DefaultIssueSelector` and overrides `addIssuesFromChangeLog()`

so the current design relies on inheritance and method overriding and the utils class would break that flow

I think adding an abstract class `BaseIssueSelector` in between `AbstractIssueSelector` and the concrete selectors to hold the shared logic could be a cleaner solution

this approach keeps `AbstractIssueSelector` as-is, `JobIssueSelector` can still override methods from the base class and avoids code duplication and the semantic issue of having `SinceLastSuccessfulBuildIssueSelector` extend `DefaultIssueSelector`

The hierarchy would be:

AbstractIssueSelector --> BaseIssueSelector --> DefaultIssueSelector
AbstractIssueSelector --> BaseIssueSelector --> SinceLastSuccessfulBuildIssueSelector

What do you think?

rantoniuk:
Let's leave it then as it is now (in AbstractIssueSelector), just make sure to cover the changes with tests accordingly ([see some missing branches](https://github.com/jenkinsci/jira-plugin/pull/746/checks?check_run_id=52503493134)).

Edit: I actually need to see if this is valid at all - see [my last comment](https://github.com/jenkinsci/jira-plugin/issues/453#issuecomment-1185705121). Allow me some time to test this on a live instance.

rantoniuk:
@ahmedanwar1 I looked at the implementation and I'm wondering what do you think about a different approach.

How about having a `sinceLastSuccessfulBuild` _flag_ in the _DefaultIssueSelector_, with a default setting of `false` (for backwards compatibility)? 
I feel that this would simplify the change, maintenance and also changes to the actual pipeline DSLs. Would that also work for you?

rantoniuk:
@ahmedanwar1 did you have a chance to look at my previous comment? Any thoughts?

ahmedanwar1:
Hi @rantoniuk 
Sorry for the late response, I’ve been quite busy over the past few months

From a design perspective i leaned toward keeping `DefaultIssueSelector` and `SinceLastSuccessfulBuildIssueSelector` as independent implementations since they represent different selection strategies

Introducing a flag in `DefaultIssueSelector` would couple multiple behaviors into a single class which could make it harder to maintain or extend if additional selector variations are added in the future

Keeping them separate helps preserve a clearer separation of concerns and avoids accumulating conditional logic in one place over time

What do you think is the best fit?

rantoniuk:
@ahmedanwar1 let's go with the flag approach then. Adding `sinceLastSuccessfulBuild` (default `false`) to `DefaultIssueSelector` keeps this to one selector, avoids the shared-logic placement question we've been going back and forth on, and stays backward compatible.

And sorry for the very delayed answer!

If you're up for reworking it that way, happy to review. Otherwise let me know and I can pick it up.

ahmedanwar1:
Thanks @rantoniuk , I'm a bit tied up at the moment so please feel free to pick it up. I really appreciate your guidance on the approach

rantoniuk:
@ahmedanwar1 just to clarify, can you rebase this to resolve the conflicts and apply the direction from https://github.com/jenkinsci/jira-plugin/pull/746#issuecomment-5369176527?

ahmedanwar1:
@rantoniuk  I had a second look and since `P4JobIssueSelector` actually depends on `DefaultIssueSelector`, introducing the flag and modifying that class would bring us back to the backward compatibility concerns we discussed before

To avoid that I think we could revisit the solution you suggested earlier (make `SinceLastSuccessfulBuildIssueSelector` extend `DefaultIssueSelector`) and treat it as the main selector (similar to how `P4JobIssueSelector` is implemented)

this would avoid adding complexity or introducing potential issues in the dependent classes by modifying the default selector
and It also preserves the existing convention where each selector has its own class rather than combining multiple behaviors into a single class

What do you think?

rantoniuk:
Yep, I agree, extending `DefaultIssueSelector` makes sense!

rantoniuk:
LGTM! Can you also update the [features.md](https://github.com/jenkinsci/jira-plugin/blob/master/docs/features.md)?

ahmedanwar1:
Sure thing! I've updated `features.md` to clarify the purpose of the new selector

PR #755, opened 2025-11-02
remove httpclient shadowing in favor of the wrapper library

rantoniuk:
Since now the wrappers are available for:
- https://github.com/jenkinsci/apache-httpcomponents-client-4-api-plugin
- https://github.com/jenkinsci/apache-httpcomponents-client-5-api-plugin

and we're using one already, I think we should probably be able to remove the class shadowing introduced in https://github.com/jenkinsci/jira-plugin/commit/abd092848b1e617ceb31b16de07b1445d34c4d03.


rantoniuk:
Closing per [ADR-0003](https://github.com/rantoniuk/jira-plugin/blob/docs/modernisation-adrs/docs/adr/0003-future-of-the-vendored-atlassian-http-client.md) (rantoniuk/jira-plugin#212) — not because the direction is wrong, but because the vendored tree isn't a pure copy of upstream: `ApacheAsyncHttpClient` reads `hudson.ProxyConfiguration` and installs a `JenkinsProxyRoutePlanner` honoring `ProxyConfiguration.getNoProxyHostPatterns()`, covered by `JiraRestServiceProxyTest`. Merging as-is would silently regress Jenkins proxy support for self-hosted/Data Center users behind a corporate proxy.

Decision for 3.x: keep and fix the fork in place (license headers, bring it under static analysis, fix the known lifecycle/timeout defects), with HttpClient 5 as the declared 4.0 target — see rantoniuk/jira-plugin#212 for the tracked issues.

Reopen/rework this along the original idea if someone establishes that `atlassian-httpclient-library` (or the httpcomponents wrapper plugins) exposes a route-planner seam the proxy config can hook into without patching internals — that would supersede this ADR.

Issue #760, opened 2025-11-27
Wrong styling of selection box es of jiraIssue and jiraReleaseVersion


ugrave:
### Jenkins and plugins versions report

<details>
  <summary>Environment</summary>

  
  ```text
Jenkins: 2.516.1
OS: Linux - 5.10.245-241.978.amzn2.x86_64
---
active-directory:2.40
analysis-model-api:13.8.0-902.v26f80296f743
ansicolor:1.0.6
antisamy-markup-formatter:173.v680e3a_b_69ff3
apache-httpcomponents-client-4-api:4.5.14-269.vfa_2321039a_83
apache-httpcomponents-client-5-api:5.5-157.vb_43ef2f6845e
asm-api:9.8-163.vb_2a_96d3f9c3c
authentication-tokens:1.144.v5ff4a_5ec5c33
authorize-project:2.0.0
aws-credentials:248.v78a_dcfc9db_ff
aws-java-sdk:1.12.780-480.v4a_0819121a_9e
aws-java-sdk-api-gateway:1.12.780-480.v4a_0819121a_9e
aws-java-sdk-autoscaling:1.12.780-480.v4a_0819121a_9e
aws-java-sdk-cloudformation:1.12.780-480.v4a_0819121a_9e
aws-java-sdk-cloudfront:1.12.780-480.v4a_0819121a_9e
aws-java-sdk-cloudwatch:1.12.780-480.v4a_0819121a_9e
aws-java-sdk-codebuild:1.12.780-480.v4a_0819121a_9e
aws-java-sdk-codedeploy:1.12.780-480.v4a_0819121a_9e
aws-java-sdk-ec2:1.12.780-480.v4a_0819121a_9e
aws-java-sdk-ecr:1.12.780-480.v4a_0819121a_9e
aws-java-sdk-ecs:1.12.780-480.v4a_0819121a_9e
aws-java-sdk-efs:1.12.780-480.v4a_0819121a_9e
aws-java-sdk-elasticbeanstalk:1.12.780-480.v4a_0819121a_9e
aws-java-sdk-elasticloadbalancingv2:1.12.780-480.v4a_0819121a_9e
aws-java-sdk-iam:1.12.780-480.v4a_0819121a_9e
aws-java-sdk-kinesis:1.12.780-480.v4a_0819121a_9e
aws-java-sdk-lambda:1.12.780-480.v4a_0819121a_9e
[...]

PR #761, opened 2025-11-27
fix: styling of select boxes for issue and version parameter

ugrave:
### Related issue

#760

### Changes

Before:
before

After:
after

### Tests

 

- [x] I have tested my changes with a Jira Cloud / Jira Server

### DIFF
--- a/src/main/java/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinition.java
+++ b/src/main/java/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinition.java
@@ -198,7 +198,7 @@ public class JiraVersionParameterDefinition extends ParameterDefinition {
                 }
             }
             if (items.isEmpty()) {
-                items.add(Messages.JiraVersionParameterDefinition_NoIssueMatchedSearch(), "");
+                items.add(Messages.JiraVersionParameterDefinition_NoVersionsMatchedSearch(), "");
             }
             return items;
         }
--- a/src/main/resources/hudson/plugins/jira/Messages.properties
+++ b/src/main/resources/hudson/plugins/jira/Messages.properties
@@ -48,4 +48,4 @@ JiraSite.threadExecutorMinimunSize = Thread Executor Size must be at least {0} (
 JiraSite.timeoutMinimunValue = Connection timeout must be at least {0}
 JiraSite.readTimeoutMinimunValue = Read timeout must be at least {0}
 JiraIssueParameterDefinition.NoIssueMatchedSearch = No issues matched the search
-JiraVersionParameterDefinition.NoIssueMatchedSearch = No version matched the search
+JiraVersionParameterDefinition.NoVersionsMatchedSearch = No version matched the search
--- a/src/test/java/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinitionTest.java
+++ b/src/test/java/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinitionTest.java
@@ -344,7 +344,7 @@ class JiraVersionParameterDefinitionTest {
             assertThat(result, hasSize(1));
             ListBoxModel.Option option = result.get(0);
             assertEquals("", option.value);
-            assertEquals(Messages.JiraVersionParameterDefinition_NoIssueMatchedSearch(), option.name);
+            assertEquals(Messages.JiraVersionParameterDefinition_NoVersionsMatchedSearch(), option.name);
             verify(job).hasPermission(Item.BUILD);
         }
 
@@ -362,7 +362,7 @@ class JiraVersionParameterDefinitionTest {
             assertThat(result, hasSize(1));
             ListBoxModel.Option option = result.get(0);
             assertEquals("", option.value);
-            assertEquals(Messages.JiraVersionParameterDefinition_NoIssueMatchedSearch(), option.name);
+            assertEquals(Messages.JiraVersionParameterDefinition_NoVersionsMatchedSearch(), option.name);
             verify(job).hasPermission(Item.BUILD);
         }
     }


src/main/resources/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinition/index.jelly
```diff
@@ -4,7 +4,7 @@
   <j:set var="escapeEntryTitleAndDescription" value="false"/>
   <f:entry title="${h.escape(it.name)}" description="${it.formattedDescription}">
     <!-- this div is required because of ParametersDefinitionProperty.java#117 -->
-    <div name="parameter">
```
rantoniuk, line 7: Can you try removing the divs altogether and see what happens? 
I think the comment above is no longer relevant, since it was introduced 14y ago and I don't see ParametersDefinitionProperty.java anymore - ideally using https://weekly.ci.jenkins.io/design-library/select/ should be enough.

ugrave: With the `f.select` its shows the selection box. But how can i now fill it with the options?
The forEach is not working in f:select.
I also try to add a `doFillValueItems` method into the `DescriptorImpl` of the `JiraIssueParameterDefinition` but there i did`t have access to the JiraIssueParameterDefinition config. At least i did`t known ho i can access it from there.
Putting the doFillValueItems method directly in the JiraIssueParameterDefinition  was not working.

rantoniuk: Here's an example how it works for JiraProject selection dropdown:
https://github.com/jenkinsci/jira-plugin/blob/984160c70cd0e54415d4ad90de18f6586d4dbf81/src/main/java/hudson/plugins/jira/JiraProjectProperty.java#L108
https://github.com/jenkinsci/jira-plugin/blob/984160c70cd0e54415d4ad90de18f6586d4dbf81/src/main/resources/hudson/plugins/jira/JiraProjectProperty/config.jelly#L3

ugrave: For testing I added added this method:
```java
    public ListBoxModel doFillValueItems() {
        ListBoxModel items = new ListBoxModel();

        items.add("Dummy 1", "dummy1");
        return items;
    }
```
to this class:
https://github.com/ugrave/jira-plugin/blob/718150cb8edbd205dee75e299b46ee62a52d84d7/src/main/java/hudson/plugins/jira/listissuesparameter/JiraIssueParameterDefinition.java#L120

But this method is not called and also the dummy value is not shown if use Run build with params.
I`m only see an empty selection box (which correct styling)

In your example the selection is in the config.jelly file. Not sure if it is a different but I have it for the parameters in the index.jelly file.

ugrave: After some investigation i found out the the select box did`t get the fillUrl attribute on the select html element.
In the network tab i can see it then makes a request to http://localhost:8080/job/Test/null. null would be normally the content of the fillUrl.
But not sure why it does not get the fillUrl attribute.

ugrave: Found the problem.
In need to set
```xml
<j:set var="instance" value="${it}"/>
<j:set var="descriptor" value="${it.descriptor}"/>
```
in the template because the `select` expect a property `instance` and `descriptor` to configure it correctly.

src/main/java/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinition.java
```diff
@@ -173,6 +176,19 @@ public static class DescriptorImpl extends ParameterDescriptor {
         public String getDisplayName() {
             return "Jira Release Version Parameter";
         }
+
+        public ListBoxModel doFillValueItems(@AncestorInPath Job<?, ?> job, @QueryParameter String name) {
```
ugrave: Do I need to check any permission here?

src/main/resources/hudson/plugins/jira/listissuesparameter/JiraIssueParameterDefinition/index.jelly
```diff
@@ -17,25 +17,10 @@ limitations under the License.
...
   <j:set var="escapeEntryTitleAndDescription" value="false"/>
-  <f:entry title="${h.escape(it.name)}" description="${it.formattedDescription}">
-    <!-- this div is required because of ParametersDefinitionProperty.java#117 -->
-    <div name="parameter">
+  <j:set var="instance" value="${it}"/>
+  <j:set var="descriptor" value="${it.descriptor}"/>
+  <f:entry field="value" title="${h.escape(it.name)}" description="${it.formattedDescription}">
       <input type="hidden" name="name" value="${it.name}"/>
-      <j:choose>
-        <j:when test="${it.issues == null or it.issues.size() == 0}">
```
ugrave, line 25: I did`t have now direct access to the list of items. Then now coming form the fill method.
How should this now be handled without fetching the issues again?

src/main/resources/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinition/index.jelly
ugrave: I tried to remove the `<div name="parameter">`. But then the selected value is not taken over and inside of the pipleine the param value is `null`. 
With the div, the selected values is available in the pipeline param.

src/main/resources/hudson/plugins/jira/listissuesparameter/JiraIssueParameterValue/value.jelly
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 <j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
   <j:set var="escapeEntryTitleAndDescription" value="false"/>
   <f:entry title="${h.escape(it.name)}">
-    <f:textbox name="${it.name}" value="${it.issue}" />
+    <j:set var="readOnlyMode" value="true"/>
+    <f:textbox name="${it.name}" value="${it.value}"/>
```
ugrave, line 21: Fixed the issue that the parameter value was not shown on the parameter overview page of a build

src/main/java/hudson/plugins/jira/listissuesparameter/JiraIssueParameterDefinition.java
```diff
@@ -22,22 +22,19 @@
 import com.atlassian.jira.rest.client.api.domain.IssueField;
 import hudson.Extension;
 import hudson.cli.CLICommand;
-import hudson.model.Job;
-import hudson.model.ParameterDefinition;
-import hudson.model.ParameterValue;
+import hudson.model.*;
```
rantoniuk: don't convert to wildcards

src/main/java/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinition.java
```diff
@@ -69,7 +67,10 @@ public ParameterValue createValue(CLICommand command, String value) throws IOExc
 
     public List<JiraVersionParameterDefinition.Result> getVersions() throws IOException, RestClientException {
         Job<?, ?> contextJob = Stapler.getCurrentRequest2().findAncestorObject(Job.class);
+        return getVersions(contextJob);
+    }
 
+    private List<JiraVersionParameterDefinition.Result> getVersions(Job<?, ?> contextJob) {
```
rantoniuk: Why is this extracted to a private method? I don't neglect it, but I don't see it being needed (e.g. for testing).

src/main/resources/hudson/plugins/jira/listissuesparameter/JiraIssueParameterDefinition/index.jelly
```diff
@@ -16,26 +16,13 @@ limitations under the License.
...
-            <j:forEach var="issue" items="${it.issues}">
-              <option value="${issue.key}">${issue.key}: ${issue.summary}</option>
-            </j:forEach>
-          </select>
-        </j:otherwise>
-      </j:choose>
-    </div>
-  </f:entry>
+    <j:set var="escapeEntryTitleAndDescription" value="false"/>
+    <j:set var="instance" value="${it}"/>
```
rantoniuk: what is the `instance` used for?

src/main/resources/hudson/plugins/jira/listissuesparameter/JiraIssueParameterDefinition/index.jelly
```diff
@@ -16,26 +16,13 @@ limitations under the License.
...
+    <j:set var="descriptor" value="${it.descriptor}"/>
+    <f:entry field="value" title="${h.escape(it.name)}" description="${it.formattedDescription}">
+        <div name="parameter">
```
rantoniuk, line 22: The initial idea was to get rid of this `<div>`, I guess it didn't work?

src/main/resources/hudson/plugins/jira/listissuesparameter/JiraIssueParameterValue/value.jelly
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 <j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
   <j:set var="escapeEntryTitleAndDescription" value="false"/>
   <f:entry title="${h.escape(it.name)}">
-    <f:textbox name="${it.name}" value="${it.issue}" />
+    <j:set var="readOnlyMode" value="true"/>
```
rantoniuk, line 20: Is this needed?

src/main/resources/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinition/index.jelly
rantoniuk: I guess you resolved that, but I still see the `<div>` in the PR, why?

src/main/resources/hudson/plugins/jira/versionparameter/JiraVersionParameterValue/value.jelly
```diff
@@ -2,6 +2,7 @@
 <j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
   <j:set var="escapeEntryTitleAndDescription" value="false"/>
   <f:entry title="${h.escape(it.name)}">
-    <f:textbox name="${it.name}" value="${it.version}" />
+    <j:set var="readOnlyMode" value="true"/>
```
rantoniuk, line 5: Why `readonly` is needed?

src/main/java/hudson/plugins/jira/listissuesparameter/JiraIssueParameterDefinition.java
```diff
@@ -122,6 +122,22 @@ public static class DescriptorImpl extends ParameterDescriptor {
         public String getDisplayName() {
             return "Jira Issue Parameter";
         }
+
+        @RequirePOST
+        public ListBoxModel doFillValueItems(@AncestorInPath Job<?, ?> job, @QueryParameter String name) {
+            ListBoxModel items = new ListBoxModel();
+            if (job.hasPermission(Item.BUILD)) {
```
rantoniuk, line 139: This is great, but that needs a test, especially since relevant permissions are involved in here.

src/main/resources/hudson/plugins/jira/versionparameter/JiraVersionParameterValue/value.jelly
ugrave: This field is shown on the build summary page.
image
Makes no sense to have it editable here.

src/main/resources/hudson/plugins/jira/listissuesparameter/JiraIssueParameterDefinition/index.jelly
ugrave: It's internally used by the `f:select`.

src/main/resources/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinition/index.jelly
ugrave: I tried it out, but then the value is not taken over into the params object.
Inside of the pipeline `params.PARAM_NAME` is then `null` without the surrounding `div`.

src/main/java/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinition.java
ugrave: I created a new private method out of it to no break the api if a change method signature of the other one.

ugrave: I put it now to public because is used it in tests now.

src/main/resources/hudson/plugins/jira/listissuesparameter/JiraIssueParameterValue/value.jelly
ugrave: see https://github.com/jenkinsci/jira-plugin/pull/761#discussion_r2577336408

src/main/resources/hudson/plugins/jira/listissuesparameter/JiraIssueParameterDefinition/index.jelly
ugrave: see https://github.com/jenkinsci/jira-plugin/pull/761#discussion_r2577350212

src/test/java/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinitionTest.java
```diff
@@ -258,4 +268,41 @@
...
+            JiraVersionParameterDefinition.Result item = new JiraVersionParameterDefinition.Result(version);
+            when(paramDef.getVersions(any())).thenReturn(List.of(item));
+
+            ListBoxModel result = uut.doFillVersionItems(job, "PARAM_NAME");
+
+            assertThat(result, hasSize(1));
+            ListBoxModel.Option option = result.get(0);
+            assertEquals("1.0.0", option.value);
+            assertEquals("1.0.0", option.name);
+            verify(job).hasPermission(Item.BUILD);
```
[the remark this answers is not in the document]
rantoniuk: missing assertTrue

src/main/resources/hudson/plugins/jira/listissuesparameter/JiraIssueParameterDefinition/index.jelly
rantoniuk: That should not be needed, see e.g. https://github.com/jenkinsci/jira-plugin/blob/master/src/main/resources/hudson/plugins/jira/JiraProjectProperty/config.jelly#L4

src/test/java/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinitionTest.java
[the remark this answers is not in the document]
ugrave: The `verify` function tests if the method was called.  Its not intended to test the return value of the `hasPermission` method in this case. 
Seems that the security scanner is not aware of this.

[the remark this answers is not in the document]
ugrave: Sees such issues can be suppressed vie `@SuppressWarnings("lgtm[jenkins/has-permission-return-value-ignored]")`
Should I added these annotations on method level?

src/main/java/hudson/plugins/jira/listissuesparameter/JiraIssueParameterDefinition.java
```diff
@@ -122,6 +131,22 @@ public static class DescriptorImpl extends ParameterDescriptor {
...
+                ParametersDefinitionProperty prop = job.getProperty(ParametersDefinitionProperty.class);
+                if (prop != null) {
+                    ParameterDefinition def = prop.getParameterDefinition(name);
+                    if (def instanceof JiraIssueParameterDefinition jiraIssueDef) {
+                        List<Result> issueValues = jiraIssueDef.getIssues(job);
+                        issueValues.forEach(it -> items.add(it.summary, it.key));
```
rantoniuk:
```suggestion
                        issueValues.forEach(it -> items.add(String.format("%s: %s",it.key,it.summary), it.key));
```

src/main/java/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinition.java
```diff
@@ -4,21 +4,19 @@
...
 import java.util.List;
 import java.util.Objects;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import net.sf.json.JSONObject;
 import org.jenkinsci.Symbol;
-import org.kohsuke.stapler.DataBoundConstructor;
-import org.kohsuke.stapler.Stapler;
-import org.kohsuke.stapler.StaplerRequest2;
+import org.kohsuke.stapler.*;
```
rantoniuk: don't replace imports with wildcards

src/main/resources/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinition/index.jelly
rantoniuk: Indeed I see the same, that's fishy, will look into that.

src/main/resources/hudson/plugins/jira/listissuesparameter/JiraIssueParameterDefinition/index.jelly
rantoniuk: Can you keep the logic for no-items found?

src/main/java/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinition.java
rantoniuk: Tests are in the same package, it shouldn't need to be public.

src/main/resources/hudson/plugins/jira/listissuesparameter/JiraIssueParameterDefinition/index.jelly
ugrave: I did't have access to the items list of the select box. So I added an entry to the ListBoxModel with the message as the Message text.
The value cannot be `null`, so i use empty string. In the `createValue` in the param definition check for empty string and return null.
In the pipeline the param has then the value `null`.
Did`t find a better way to handle it.

ugrave: This is how it looks like:
image

ugrave: Maybe it also nice to have a parameter to say if the parameter is required. 
Then we could prevent pressing the build button without a valid issue selected.
Ex the git-parameter has it.
Something for a new PR.

src/main/java/hudson/plugins/jira/listissuesparameter/JiraIssueParameterDefinition.java
```diff
@@ -123,6 +132,25 @@ public static class DescriptorImpl extends ParameterDescriptor {
...
+            if (job.hasPermission(Item.BUILD)) {
+                ParametersDefinitionProperty prop = job.getProperty(ParametersDefinitionProperty.class);
+                if (prop != null) {
+                    ParameterDefinition def = prop.getParameterDefinition(name);
+                    if (def instanceof JiraIssueParameterDefinition jiraIssueDef) {
+                        List<Result> issueValues = jiraIssueDef.getIssues(job);
+                        issueValues.forEach(it -> items.add(it.key + ": " + it.summary, it.key));
+                    }
+                }
+                if (items.isEmpty()) {
```
rantoniuk: Move outside the last `if`

src/main/java/hudson/plugins/jira/versionparameter/JiraVersionParameterDefinition.java
```diff
@@ -173,6 +183,25 @@ public static class DescriptorImpl extends ParameterDescriptor {
...
+            if (job.hasPermission(Item.BUILD)) {
+                ParametersDefinitionProperty prop = job.getProperty(ParametersDefinitionProperty.class);
+                if (prop != null) {
+                    ParameterDefinition def = prop.getParameterDefinition(name);
+                    if (def instanceof JiraVersionParameterDefinition jiraVersionDef) {
+                        List<JiraVersionParameterDefinition.Result> issueValues = jiraVersionDef.getVersions(job);
+                        issueValues.forEach(it -> items.add(it.name));
+                    }
+                }
+                if (items.isEmpty()) {
```
rantoniuk: Move outside last `if`

src/main/resources/hudson/plugins/jira/Messages.properties
```diff
@@ -47,3 +47,5 @@ ErrorCommentingIssues=[Jira] Could not comment on some issues: {0}
 JiraSite.threadExecutorMinimunSize = Thread Executor Size must be at least {0} (higher values are recommended)
 JiraSite.timeoutMinimunValue = Connection timeout must be at least {0}
 JiraSite.readTimeoutMinimunValue = Read timeout must be at least {0}
+JiraIssueParameterDefinition.NoIssueMatchedSearch = No issues matched the search
+JiraVersionParameterDefinition.NoIssueMatchedSearch = No version matched the search
```
rantoniuk:
```suggestion
JiraVersionParameterDefinition.NoVersionsMatchedSearch = No version matched the search
```

rantoniuk:
Please don't use force-pushing when adressing review comments because it makes it harder to review changes - the PR will be squashed anyway during merge.

Issue #758, opened 2025-11-23
[contributing] Checkstyle setup for IntelliJ


zbynek:
The Checkstyle setup in .idea folder defines Google style as active, so devs who use Checkstyle IntelliJ plugin se hundreds of warnings because "expected" indentation is 2 and real one is 4 spaces.

I would suggest either removing the file from Git if maintainers do not use this plugin or changing it in a way that works for Jenkins checkstyle config.

https://github.com/jenkinsci/jira-plugin/blob/master/.idea/checkstyle-idea.xml

rantoniuk:
> works for Jenkins checkstyle config

Which config do you have in mind?

zbynek:
Maybe https://raw.githubusercontent.com/jenkinsci/jenkins/refs/tags/jenkins-2.516.3/src/checkstyle/checkstyle-configuration.xml ?

rantoniuk:
I'm ok with removing this configuration altogether, but at the same time I would like to make sure that the IDE formatting (on-demand or on-save) is compliant [with the linters coming from the plugin-pom](https://github.com/jenkinsci/jenkins/blob/2ba827b6c523091428b706710f4f0fcf55d8e06d/CONTRIBUTING.md?plain=1#L96). I don't use IntelliJ Idea anymore, but since you've opened this issue, I'm assuming you're still using it? 

If removing the `.idea` folder resolves the issue and auto-formatted (in IDE) code is compliant with `mvn spotless:check`, then let's go for that.

zbynek:
The list of linters you linked above is relevant for Jenkins core, the plugin POM does not bring any Checkstyle config. If this plugin were to use core's Checkstyle config, it would require code style changes proposed in  #762 .

> If removing the .idea folder resolves the issue and auto-formatted (in IDE) code is compliant with mvn spotless:check, then let's go for that.

That sounds like a good solution to me.

PR #762, opened 2025-11-30
Code style improvements

zbynek:
Related to #758 

Some code style changes to make the code style more aligned with Jenkins core.


src/main/java/hudson/plugins/jira/Updater.java
```diff
@@ -52,12 +52,11 @@ class Updater {
      */
     public static boolean debug = false;
 
-    public Updater(SCM scm) {
+    Updater(SCM scm) {
         this(scm, new ArrayList<>());
     }
 
-    public Updater(SCM scm, List<String> labels) {
```
rantoniuk, line 59: Package visibility change - why?

src/main/java/hudson/plugins/jira/Updater.java
```diff
@@ -52,12 +52,11 @@ class Updater {
...
-        super();
```
rantoniuk, line 60: That's not a code style change, why?

src/main/java/hudson/plugins/jira/selector/JqlIssueSelector.java
```diff
@@ -25,7 +25,6 @@ public class JqlIssueSelector extends AbstractIssueSelector {
 
     @DataBoundConstructor
     public JqlIssueSelector(String jql) {
-        super();
```
rantoniuk, line 28: Why?

zbynek: There is a checkstyle rule for not calling `super()` explicitly if it doesn't do anything (the defualt construstructor is implicitly called).

https://checkstyle.sourceforge.io/checks/coding/avoidnoargumentsuperconstructorcall.html

src/main/java/hudson/plugins/jira/Updater.java
zbynek: Since the class itself is not public, the constructor cannot be effectively public either.
The rule that triggered this: https://checkstyle.sourceforge.io/checks/modifier/redundantmodifier.html

zbynek:
@rantoniuk these changes will align the style with Jenkins core. They will not help for [checkstyle-idea.xml](https://github.com/jenkinsci/jira-plugin/blob/master/.idea/checkstyle-idea.xml) which is based on Google code style (requires 2 space indentation instead of 4 spaces).

rantoniuk:
> @rantoniuk these changes will align the style with Jenkins core. They will not help for [checkstyle-idea.xml](https://github.com/jenkinsci/jira-plugin/blob/master/.idea/checkstyle-idea.xml) which is based on Google code style (requires 2 space indentation instead of 4 spaces)

Then I guess you still want to remove that file as part of this PR as discussed earlier?

PR #1197, opened 2026-08-22
fix(http): apply connect and read timeouts to Jira HTTP client

### LINKED ISSUE #215: The "Read timeout" setting has no effect, and connect timeout is never applied
**Type:** Bug

Part of #212 (Epic 2 — HTTP stack correctness).

rantoniuk:
## Summary
- `JiraSite.getHttpClientOptions()` never called `setConnectionTimeout()`, so the connect timeout configured in the Jira site settings was always overridden by the vendored client's hardcoded 5s default.
- The "Read timeout" value was wired into `setRequestTimeout()`, which nothing in the vendored Apache HTTP client actually reads (verified no call sites of `getRequestTimeout()` anywhere in `src/main`), making that setting a complete no-op.
- The socket/read timeout that *is* honored (`setSocketTimeout`, read by `ApacheAsyncHttpClient`) was instead being fed the connect-timeout value.

Fixed by wiring `timeout -> setConnectionTimeout` and `readTimeout -> setSocketTimeout`.

Fixes https://github.com/rantoniuk/jira-plugin/issues/215

## Test plan
- [x] Added `JiraSiteTest.httpClientOptionsApplyConnectAndReadTimeouts`, verified it fails against the old code (`expected: <7000> but was: <5000>`) and passes with the fix.
- [x] `JiraSiteTest`, `JiraSessionTest`, `JiraRestServiceTest`, `JiraRestServiceGetMyselfTest`, `JiraRestServiceProxyTest` all green (60 tests).


rantoniuk:
Superseded by #1200, which includes this same timeout fix (rebased on current master) alongside the other JiraSite fixes it was originally split from in #1198.

PR #231, opened 2020-04-21
Reenable full build on TravisCI

warden:
As JenkinsCI is sometimes failing and to have the parity with it in order not to confuse contributors, reenable normal full build on TravisCI too.
Replaces #223 
Reasons:
https://issues.jenkins-ci.org/browse/INFRA-2569
https://issues.jenkins-ci.org/browse/INFRA-2548
https://issues.jenkins-ci.org/browse/INFRA-2544
https://issues.jenkins-ci.org/browse/INFRA-2393

PR #1198, opened 2026-08-23
fix: resolve every bug-labelled issue in the 3.x milestone

rantoniuk:
### Epic 2 — HTTP stack correctness

| Commit | Issue | Defect |
|---|---|---|
| `ad05f98` | #215 | `setRequestTimeout` is never read by the client, and connect timeout was never configured — "Read timeout" had no HTTP-layer effect and connect timeout was stuck at the library's 5 s default |
| `0cf1dbb` | #214 | `DefaultHttpClientFactory` cached its client in a **non-volatile `static`**, read outside any lock, written under `synchronized (this)` — a per-instance monitor guarding static state. The first site to build a client pinned its `HttpClientOptions` onto every other site until restart |
| `b02d3b6` | #216 | The callback executor was `static` but sized from an **instance** field, and `ApacheAsyncHttpClient.destroy()` calls `shutdown()` on it — closing one site's client disabled Jira for every site on the controller |

Per [ADR-0006], the per-request client build in `getPromiseHttpAsyncClient` is **deliberately left
alone**.

### Epic 3 — code health

| Commit | Issue | Defect |
|---|---|---|
| `20741f3` | #225 | The session was cached with **no key**, while credentials resolve per `Item`. |
| `195fc29` | #226 | "No session yet" was stored as `Optional.empty()` and served as "issue does not exist"; cache was unbounded and `expireAfterAccess`, so the annotator's own re-reads kept renewing poisoned entries |
| `d6f2857` | #227 | Project list fetched once per JVM (its own `FIXME` said so); lock timeout, `RestClientException` and interrupt all collapsed into a silent empty set, which makes every changelog issue link disappear |
| `2815fc4` | #229 | `sites.add()` on `Collections.emptyList()` — threw `UnsupportedOperationException` on every freshly constructed folder property. There was no code path on which it worked |
| `b93802f` | #230 | `jiraIssueSelector()` with no selector → bare `NullPointerException` |
| `2e03212` | #231 | `jiraSearch` dereferenced the `@Nullable` `JiraSite.get(...)` on the next line → bare `NullPointerException` |
| `4af0051` | #232 | `startDate` participates in `equals` but not `hashCode` |
| `9b733b0` | #233 | Three identical `GET /issue/{key}` per workflow transition |
| `161bef4` | #234 | A failed `SystemCredentialsProvider.save()` was logged and then reported as success, leaving the site pointing at a credential that only exists in memory |
| `4caca77` | #235 | Issue keys persisted twice; two constructors left the form-bound field null, so the config page rendered empty and the next save wiped them |

### Epics 4 and 5

| Commit | Issue | Defect |
|---|---|---|
| `385217f` | #248 | The "offline" WireMock suite downloaded a 2.3 MB OpenAPI spec in a static initialiser |
| `d282e4d` | #258 | Every field id was prefixed `customfield_`, so `labels` and every other built-in field was unreachable from the step |

### Found while working through the above

| Commit | Issue | Defect |
|---|---|---|
| `a749755` | #276 | `JiraIssueField.compareTo` delegates to itself → `StackOverflowError` |
| `5ea73fb` | #277 | `JiraVersion.compareTo` dereferences the null release dates its own constructors produce |

## Worth a reviewer's attention

- **#230 changes behaviour**: `jiraIssueSelector()` with no selector now falls back to
  `DefaultIssueSelector` instead of failing. Safe here because the step only *reads* issue ids —
  `jiraUpdateIssueField` still aborts, since guessing which issues to modify is another matter.
- **#234 now throws** `FormException` instead of silently succeeding. `JiraSite.readResolve` catches
  it and loads the site without credentials — a site must never fail to deserialise.
- **#258 changes `prepareFieldId` semantics**: only a bare number is prefixed now, so an input like
  `field_10100` is passed through rather than becoming `customfield_field_10100`. That form was never
  documented, and the existing test asserting it has been updated.
- **#248 adds a 165 KB test resource** (trimmed from 2.3 MB) plus `tools/trim-jira-openapi-spec.mjs`
  to regenerate it. Validating against a path the trimmed copy lacks fails loudly, so trimming cannot
  quietly become "validating nothing".
- **Only one public method is removed**: `doCheckField_id`, renamed to `doCheckFieldId`. It was dead —
  the form binds `field="fieldId"`, so Stapler never called it.

## Scope notes

- #258 is deliberately **minimal** and does not attempt the structured/multi-value rewrite that the
  epic pairs it with — only `labels` (a plain string array) is handled; `components`/`fixVersions`
  need arrays of objects and are left out.
- `JiraSite` is **not** restructured here; extracting client construction and issue-link caching stays
  a separate task.
- Happy to split this into per-epic PRs if that reviews better against the diff-scoped quality gate.

[ADR-0006]: https://github.com/rantoniuk/jira-plugin/blob/docs/modernisation-adrs/docs/adr/0006-http-client-lifecycle-and-jenkins-60536.md


rantoniuk:
Splitting this into per-issue PRs for a diff-scoped review, as suggested in the description. Closing
this in favor of:

- #1199 — HTTP client caching per factory
- #1200 — JiraSite connection, caching and session-lifecycle fixes
- #1201 — JiraVersion equals/hashCode/compareTo
- #1202 — Jira site in a folder
- #1203 — jiraIssueSelector fallback
- #1204 — jiraSearch missing-site handling
- #1205 — Explicit issue selector persistence
- #1206 — Offline WireMock spec validation
- #1207 — Built-in fields in jiraUpdateIssueField
- #1208 — JiraIssueField.compareTo

Same commits, same tests, just regrouped so each PR's diff maps to one Sonar quality-gate evaluation
instead of one shared across the whole batch.

Issue #1216, opened 2026-09-01
All sites from glogal configuration cannot be removed


damianszczepanik:
### Jenkins and plugins versions report

Global configuration for Jira plugin allows to add or remove sites. As long as there is at least one site everything goes fine. However if user tries to delete last site and leave whole configuration empty save operation does not remove last site section.

### What Operating System are you using (both controller, and any agents involved in the problem)?

Jenkins 2.568.2, most recent LTS

### Reproduction steps

1. Add site for Jira plugin global configuration
2. Save changes
3. Delete site added in first step
4. Save changes
5. Open global configuration again -> site was not removed

### Expected Results

List is cleared

### Actual Results

Problem happens because if [`sites`](https://github.com/jenkinsci/jira-plugin/blob/4202a5dd766e0321185d2730f74f4f22d2c59442/src/main/java/hudson/plugins/jira/JiraGlobalConfiguration.java#L21C35-L21C61) collection is empty then [`setSites()`](https://github.com/jenkinsci/jira-plugin/blob/4202a5dd766e0321185d2730f74f4f22d2c59442/src/main/java/hudson/plugins/jira/JiraGlobalConfiguration.java#L32) method is not invoked and the list is not cleared

rantoniuk:
Thanks for reporting the bug - how common is that scenario? Is this limiting your usage in any way?
Also, please add the versions report as instructed in the template.

damianszczepanik:
Honestly I found that issue because I wanted to implement similar solution and then found that limitation.

I can prepare PR that eliminates this issue

rantoniuk:
> I wanted to implement similar solution and then found that limitation.

That doesn't answer my question though - what _solution_ you wanted to implement / why you needed to implement it? I'm trying to understand the usecase/scenario in which the user would hit that.

In other words, why would you want to _remove_ the last global Jira site when it was already defined?

damianszczepanik:
True , it is not problematic to me as I don't use that feature. 

Well if user doesn't want to have this configured because you he doesn't use Jira anymore then it should be able to drop unused or invalid configuration. Reasons for deleting invalid unused or test configuration may be plenty.

rantoniuk:
Got it, thanks - sure, PRs are welcome!

PR #1217, opened 2026-09-06
fix(site): global configuration did not save during site removal

damianszczepanik:
### Related issue

Issue #1216
References:
- https://github.com/jenkinsci/nuget-plugin/blob/14daa8809c7ce757d3312cecd3f56c438604962a/src/main/java/org/jenkinsci/plugins/nuget/NugetGlobalConfiguration.java#L46
- https://github.com/jenkinsci/sonarqube-plugin/blob/c4b955e3f78b43cf08686fbb7910e8ff469e7ab1/src/main/java/hudson/plugins/sonar/SonarGlobalConfiguration.java#L144

### Changes

Scenario described in #1216

### Tests

Tested on Jenkins 2.586.3 starting from last release and checking how `save()` works after plugin upgrade

- [ ] I have verified that the Code Coverage is not lower than before / that all the changes are covered as needed


damianszczepanik:
Also wondering if https://github.com/damianszczepanik/jira-plugin/commit/f1ee49837c85071eba0aeabdeb66ce689e3eec89 can be reverted as `save()` is now present in `configure()` method. Not sure about REST API that may not call `configure()` on update

rantoniuk:
Can you add respective tests please?

damianszczepanik:
> Can you add respective tests please?

That will be hard to make a good test and valuable one.
The issue is that for test I would have to mock input for configure() method and the trigger as this method is a result of saving process driven by Jenkins. Then best for the test is to make sure that super method save() is also called. In that situation almost everything needs to be mocked what makes this test useless and hard to maintain. 

But maybe you have idea how to write good test for this change.

rantoniuk:
There is a way to test it without mocks: drive the real form with HtmlUnit. Same idea as
[`ChineseWorkdayGlobalConfigurationUiTest#deletingCalendarRemovesItFromSavedConfiguration`](https://github.com/jenkinsci/chinese-workday-plugin/blob/8bf27dd1d6dee3bd8fd9c41390b3186a0a001d91/src/test/java/io/jenkins/plugins/chinese_workday/config/ChineseWorkdayGlobalConfigurationUiTest.java#L46-L74),
which covers exactly this repeatable-delete case:

```java
JiraGlobalConfiguration.get().setSites(List.of(new JiraSite("https://issues.jenkins.io/")));

HtmlForm form = r.createWebClient().goTo("configure").getFormByName("config");
form.<HtmlButton>querySelector("button.repeatable-delete").click();
r.submit(form);

assertThat(JiraGlobalConfiguration.get().getSites(), is(empty()));
```

`JenkinsRule.getButtonByCaption(form, "Delete")` will not find the button, since
  [`repeatableDeleteButton.jelly`](https://github.com/jenkinsci/jenkins/blob/jenkins-2.492.3/core/src/main/resources/lib/form/repeatableDeleteButton.jelly#L35) renders an icon plus a tooltip and no text. Hence the CSS class.

While you are in there:
[`GlobalConfiguration.configure()` already calls `req.bindJSON(this, json)`](https://github.com/jenkinsci/jenkins/blob/jenkins-2.492.3/core/src/main/java/jenkins/model/GlobalConfiguration.java#L85-L88),
so the current version binds the form three times and saves up to three times. This is enough:

```java
@Override
public boolean configure(StaplerRequest2 req, JSONObject formData) throws FormException {
    setSites(req.bindJSONToList(JiraSite.class, formData.get("sites")));
    return true;
}
```

rantoniuk:
Thanks a lot for your contribution! 🥳

